PR #93, opened 2024-12-15
Make cbor crate no_std compatible

horazont:
Thankfully, all this needed was a `#![no_std]` and a bunch of imports from the `alloc` crate (shipped with Rust).

The advantage of being `no_std` is that projects targeting embedded platforms without support for the full standard library can make use of it. Due to the alloc dependency, they will still have to provide a dynamic alloctor though.

Fixing this, as far as reasonable, needs more work.

### DIFF
--- a/cbor/src/decode.rs
+++ b/cbor/src/decode.rs
@@ -1,5 +1,10 @@
+use alloc::{
+    format,
+    string::{String, ToString},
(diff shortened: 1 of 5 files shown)


ricktaylor:
It's a great first start - thanks!

Issue #142, opened 2025-08-03
cbor: allow parsing metadata of a Value before the full data is available


horazont:
## Description

`try_parse_value` requires that the entire value is available in the buffer passed to it. This is not always desirable.

## Use case

When implementing the MTCP convergence layer, it would be useful to know the length of the bundle data which is about to be received before the reception is over. That way, the necessary memory can be preallocated or measures can be taken if the size would surpass the system's capacity.

## Proposal

I have two ideas on how to proceed from here.

### Option 1: Remove length check from `parse_data_minor`

That way, for `Value::Bytes` (only), it would be possible to parse an incomplete value. This would not gain anything for all the other types, because `Value::Text` requires to have the entire buffer available for the `from_utf8` call (and it would indeed panic in case of incomplete data, so it would need fixing there, too).

### Option 2: Introduce `ValueHead` enum

```rust
pub enum ValueHead {
    NegativeInteger(u64),
    UnsignedInteger(u64),
    Bytes {
        tags: Vec<u64>,
        range: Range<usize>,
    },
    Text {
        tags: Vec<u64>,
        range: Range<usize>,
    },
    // ..
}
```

This enum would be generated by a to-be-written `try_parse_value_head` function. Values which have known and small size (say, anything up to the size of an encoded u64) could be parsed straight away and included as variants of the `ValueHead` enum, like showcased with the `NegativeInteger` and `UnsignedInteger` variants.

---

Thoughts on this?

I personally think Option 2 is cleaner, even though it still feels clunky. Might become less clunky if it's possible to make a `impl ValueHead { fn complete(self, buf: &[u8]) -> Result<Value> }`, which naturally completes the parsing.

ricktaylor:
Hmm... an interesting thought. 

I agree your Option 1 is not great, and introduces panics.   I can where you're going with Option 2, but I'm not convinced it's the right approach, and I'm not convinced that low-level hardy_cbor is the right place to implement this logic in the first place.

My reasoning is:
1. From a 'pure rust' perspective, the parsing logic follows a deserialize pattern, similar to other deserializing crates e.g. serde, and is built around the contract of "Can I parse a type T from the bytes provided.".  Only in parsers for flexible text formats such as winnow, does one start to have APIs where partial results (e.g. "Need more") are valid results - and I'd like to keep the existing library as simple as possible.  Restarting parsers starts to require 'parser state' to be maintained somewhere, and that feels too heavy.

[...]

PR #146, opened 2025-08-05
Rework CBOR parser to handle 'push' semantics

ricktaylor:
As per #142, rework the Errors to provide some feedback when a few more bytes are needed.

Also, refactored some of the existing code to use some more modern Rust idioms

### DIFF
--- a/cbor/src/decode.rs
+++ b/cbor/src/decode.rs
@@ -145,13 +145,15 @@ fn parse_tags(data: &[u8]) -> Result<(Vec<u64>, bool, usize), Error> {
     let mut tags = Vec::new();
     let mut offset = 0;
     let mut shortest = true;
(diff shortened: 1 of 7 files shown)


cbor/src/decode.rs
```diff
@@ -234,47 +239,48 @@ where
     E: From<Error>,
 {
     let (tags, mut shortest, mut offset) = parse_tags(data)?;
-    if offset >= data.len() {
+    let Some(marker) = data.get(offset) else {
         if !tags.is_empty() {
             return Err(Error::JustTags.into());
```
ricktaylor, line 244: This should be `NeedMoreData` not `JustTags` perhaps `JustTags` should go?

ricktaylor: Done!

PR #228, opened 2025-09-07
Implement macros for cbor encode and decode

robot-head:


cbor/src/encode.rs
```diff
@@ -615,85 +603,99 @@ where
...
+    (emit_array_i16, i16),
+    (emit_array_i32, i32),
+    (emit_array_i64, i64),
+    (emit_array_isize, isize),
+    (emit_array_f16, half::f16),
+    (emit_array_f32, f32),
+    (emit_array_f64, f64),
+    (emit_array_bool, bool),
+    (emit_array_string, String)
+);
```
ricktaylor, line 698: If we have `emit_array_oneshot<T>` above, do we need this macro set, surely type coercion should handle the detail?

cbor/src/encode.rs
```diff
@@ -158,6 +158,19 @@ impl Encoder {
...
+    {
+        let count = values.len();
+        let mut a = Array::new(self, Some(count));
+
+        for value in values {
+            a.emit(value);
+        }
+        a.end()
+    }
+
```
ricktaylor, line 173: I'm just wondering if we should change this (again)...

How about we try something like:

`impl<T> ToCbor for [T]
where
    T: ToCbor` ?

cbor/src/encode.rs
```diff
@@ -2,7 +2,9 @@ use super::*;
 use core::ops::Range;
 
 pub trait ToCbor {
```
robot-head, line 4: TryToCbor?

ricktaylor: I'm not keen on TryToCbor as that implies it can fail to me, and it can't. The return value is usually (), but sometimes some additional information, so although it has a Result type, it's not a true Try

PR #277, opened 2025-11-17
Feat/cli/add crc and flags

### LINKED ISSUE #276: Add 'flags' and CRC type options to `bundle create` command
Because they are missing from the args and available to the underlying API

theotchlx:
## What does this do?

This patchset adds two CLI arguments to the `bundle create` command, for CRC type and for Bundle processing control flags.

## How?

These are custom types and clap wants to parse the values into these types, so instead of implementing traits like FromStr to the types (in the underlying API modules), I decided to write value_parser functions for clap in the CLI module.

Take a look at the commit messages for more details.

---

Closes #276

### DIFF
--- a/bpv7/tools/create.rs
+++ b/bpv7/tools/create.rs
@@ -1,6 +1,46 @@
 use super::*;
+use hardy_bpv7::bundle::Flags;
 use hardy_bpv7::crc::CrcType;
 use hardy_bpv7::eid::Eid;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
+enum ArgFlags {
+    /// Specify bundle is a fragment
+    #[value(name = "isfrag")]
+    IsFragment = 1 << 0,
+
+    /// Specify ADU is an administrative record
+    #[value(name = "isadm")]
+    IsAdminRecord = 1 << 1,
+
+    /// Require bundle to not be fragmented
+    #[value(name = "nofrag")]
+    DoNotFragment = 1 << 2,
+
+    /// Request acknowledgement by application
+    #[value(name = "ack")]
+    AppAckRequested = 1 << 5,
+
+    /// Request status time in status reports
+    #[value(name = "time")]
+    ReportStatusTime = 1 << 6,
+
+    /// Request reception status reports
+    #[value(name = "rcv")]
+    ReceiptReportRequested = 1 << 14,
+
+    /// Request forwarding status reports
+    #[value(name = "fwd")]
+    ForwardReportRequested = 1 << 16,
+
+    /// Request delivery status reports
+    #[value(name = "dlv")]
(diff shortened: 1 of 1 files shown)


theotchlx:
Hi Rick, seemed easy enough and I was too bored to sleep. Hope you can find a use for my code
Cheers

ricktaylor:
Amazing stuff Theo.  Thank you!

I'm going to be picky, and wonder if we can set the flags by using a text representation, like you've done for CRC type, just so people don't have to look in RFC9171 to work out 0xC is what they want...

(I can do it myself, but neck deep in fixing the BPSec work right now)

theotchlx:
My pleasure, no worries. Be as picky as you can!

I added a commit that lets you do stuff like `bundle create ... -f 0x8BADF00D -f is_admin_record -f do_not_fragment`, here accumulating the hex value with named flags (names taken from the Flags struct).

Hope this is what you had in mind :-)

ricktaylor:
Yeah - that feels much more idiomatic, thank you.

Can you do me one last favour?  Can you just check the help text when you run `bundle create help` gives the user some indication of the valid values?  I'm not sure what clap does with the enum, but I'm pretty sure you will need to add some helper attributes to the crc_type field.

theotchlx:
> Yeah - that feels much more idiomatic, thank you.
> 
> Can you do me one last favour? Can you just check the help text when you run `bundle create help` gives the user some indication of the valid values? I'm not sure what clap does with the enum, but I'm pretty sure you will need to add some helper attributes to the crc_type field.

Ah right, good catch!

Now it shows:

```
$ cargo run --bin bundle -- create --help
Create a new bundle with payload

Usage: bundle create [OPTIONS] --source <SOURCE> --destination <DESTINATION> --payload <PAYLOAD>

Options: 
  [...]

  -f, --flags <FLAGS>
          One or more bundle processing control flags, seperated by ','

          Possible values:
          - isfrag: Specify bundle is a fragment
          - isadm:  Specify ADU is an administrative record
          - nofrag: Require bundle to not be fragmented
          - ack:    Request acknowledgement by application
          - time:   Request status time in status reports
          - rcv:    Request reception status reports 
          - fwd:    Request forwarding status reports
          - dlv:    Request delivery status reports
          - del:    Request deletion status reports

  -c, --crc-type <CRC_TYPE>
          The CRC type of the bundle

          Possible values:
          - none:  no Cyclic Redundancy Check (CRC)
          - crc16: Standard X-25 CRC-16 [aliases: crc16_x25, 16]
          - crc32: Standard CRC32C (Castagnoli) CRC-32 [aliases: crc32_castagnoli, 32]

  [...]
```

Cool!

I added the aliases in the doc comments, otherwise clap doesn't show them. Seems to be a current limitation for ValueEnums.

Oh and now we got rid of both string parsing functions. Splendid.

PR #284, opened 2025-11-20
Bpa-server: Make log_level optional with default value + doc update

### LINKED ISSUE #283: Missing log_level field in minimal configuration example in README of bpa-server
Hey @ricktaylor, While exploring issue #272, I copied the minimal configuration example from `bpa-server/README.md` and the code panicked with the following error:

```
thread 'main' panicked at bpa-server/src/config.rs:170:10:
Failed to parse configuration: missing configuration field "log_level"
```

The minimal configuration example in `bpa-server/README.md` omits the `log_level` field, which is required by the configuration parser. When developpers copy this example, they get a startup error.

## Proposed Solutions

Two possible approaches to fix this issue:

1. **Update the documentation**: Add the `log_level` field to the minimal example in `bpa-server/README.md` (lines 40-47) to be consistent with `example_config.toml`

2. **Define a default value in the code**: Add `#[serde(default = "default_log_level")]` to the `log_level` field in `bpa-server/src/config.rs` to make this field optional with a default value

Both solutions can be combined for a better developper experience.

charley-geoffroy:
This PR addresses issue #283  which reported that copying the minimal configuration example from bpa-server/README.md would cause the program to panic with the following error:

```
thread 'main' panicked at bpa-server/src/config.rs:170:10:
Failed to parse configuration: missing configuration field "log_level"
```

The minimal configuration example was missing the `log_level` field, which is required by the configuration parser.

## Reproduce bug

To reproduce the original bug:

1. Copy the minimal configuration example from `README.md` (before this fix)
2. Save into `example_config.toml`
3. Run: `hardy-bpa-server --config example_config.toml`
4. The program panics with: `missing configuration field "log_level"`

After this fix, the same configuration file works correctly, with or without using `"info"` as the default log level.

## Other changes

- Added `static_routes_example` file: This example file facilitates onboarding for new contributors by providing a clear template for static route definitions. 

---

Closes #283

### DIFF
--- a/bpa-server/src/grpc/mod.rs
+++ b/bpa-server/src/grpc/mod.rs
@@ -21,7 +21,6 @@ pub struct Config {
         + time::Duration::nanoseconds(t.nanos.into()))
 }*/
 
-#[cfg(feature = "grpc")]
 fn to_timestamp(t: OffsetDateTime) -> prost_types::Timestamp {
     let t = t - OffsetDateTime::UNIX_EPOCH;
     prost_types::Timestamp {


bpa-server/src/config.rs
```diff
@@ -30,9 +34,10 @@ pub enum BundleStorage {
     // S3(S3Config),
 }
 
-#[derive(Serialize, Deserialize, Debug, Default)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct Config {
     // Logging level
+    #[serde(default = "default_log_level")]
     pub log_level: String,
```
ricktaylor: I'm wondering if we ought to be smarter here (as this is very old code).

Should the type actually be: `tracing::Level` not a string, and then we can implement a custom `serialize` and `deserialize` function using the `tracing::Level::FromStr` implementation?

It might be nicer than what we currently do in `start_logging` in main.rs

charley-geoffroy: I completely agree! I guess type-safe approach with `tracing::Level` (or `LevelFilter`) would be much better and more elegant than just parsing strings at runtime (and make the code cleaner).

Should this be the topic of another issue? If so, should I remove "default_log_level" and create new issue for it ?

ricktaylor:
All merged - thanks Charley!

PR #388, opened 2026-02-25
feat(bpa): persist status reports bundles until service reconnects

### LINKED ISSUE #377: Status reports dropped when originating application not registered
## Problem

  When a status report arrives at the BPA admin endpoint, the BPA attempts to notify the originating application by looking up a service matching `bundle_id.source`. However, for Applications (as opposed to low-level Services), the Report-To EID is the BPA admin endpoint, not the application itself.

  If the originating application isn't currently registered (e.g., disconnected, not yet started), the status report is silently dropped in `dispatcher/admin.rs:68`.

  This means applications that reconnect after sending bundles never receive their status reports.

  ## Design

  Add a new `BundleStatus::WaitingForService { source: Eid }` variant that queues status report bundles until the originating service registers.

  **Flow:**
  1. Status report arrives at admin endpoint
  2. Service lookup by `bundle_id.source` fails
  3. Set `bundle.metadata.status = WaitingForService { source }`
  4. Persist bundle - survives restarts
  5. When service registers, trigger `poll_service_waiting(&eid)`
  6. Matching bundles re-processed and status reports delivered

  **Key properties:**
  - Reuses existing bundle storage and metadata persistence
  - Reaper handles expiry - stale reports dropped when bundle lifetime expires
  - No spinning on route changes - only triggers when matching service registers
  - SQLite: status_code=5 with source EID in status_param3, no schema migration needed

  ## Files to Change

  - `bpa/src/metadata.rs` - Add `WaitingForService { source: Eid }` variant
  - `bpa/src/dispatcher/admin.rs` - Use new status when service not found
  - `bpa/src/services/registry.rs` - Call `poll_service_waiting()` on registration
  - `bpa/src/dispatcher/mod.rs` - Add `poll_service_waiting()` method
  - `bpa/src/storage/mod.rs` - Add `get_waiting_for_service()` trait method
  - `bpa/src/dispatcher/restart.rs` - Handle `WaitingForService` in recovery
  - `sqlite-storage/src/storage.rs` - Implement status_code=5 encoding and query

  ## References

  - Design details: `TODO.md` section 1.7
  - Current TODO: `bpa/src/dispatcher/admin.rs:43`
  - Bundle state machine: `bpa/docs/bundle_state_machine_design.md`

sylvain-pierrot:
## Summary (resolve #377)

When a status report arrives at the BPA admin endpoint and the originating application (or service) is not registered, the report is no longer dropped. The BPA stores it in `WaitingForService { source }`, persists and watches it, and re-processes when that service registers. The reaper cleans up expired ones as for any other bundle.

## Notes

- **`poll_service_waiting` instead of `get_waiting_for_service`**  
  There can be multiple bundles waiting for the same service, so the implementation streams them via `poll_service_waiting(source, tx)` rather than a getter.

- **Admin:** metadata is updated only when `bundle.metadata.status != desired` to avoid a redundant write.

- **SQLite:** `status_code = 5`, `status_param3 = source.to_string()`. No schema change.

## Remarks (for discussion)

- **State machine, readability and naming**  
  Bundle status is changed in several places (dispatch, admin, channel, adu_reassembly, metadata_mem) with no single place in code that encodes when we transition state. The machine is implemented across many small steps (receive → ingest → process_bundle, plus admin, forward, poll loops), so logic feels spread. Hard to see the state machine from the code alone. At least, maybe we should consider a helper (or small API) for all state transitions so “where do we change bundle state?” has a clear answer in code.

## Files touched

- `bpa/src/metadata.rs` — `BundleStatus::WaitingForService { source }`
- `bpa/src/dispatcher/admin.rs` — set status and `watch_bundle` when service not found
- `bpa/src/services/registry.rs` — call `poll_service_waiting` after registration
- `bpa/src/dispatcher/mod.rs` — `poll_service_waiting` (channel + load + ingest)
- `bpa/src/storage/mod.rs` — `MetadataStorage::poll_service_waiting`
- `bpa/src/storage/store.rs` — forward to backend
- `bpa/src/storage/metadata_mem.rs` — in-memory impl
- `bpa/src/dispatcher/restart.rs` — re-ingest on recovery for `WaitingForService`
- `sqlite-storage/src/storage.rs` — status code 5, query, `poll_service_waiting` impl

### DIFF
--- a/bpa/src/storage/metadata_mem.rs
+++ b/bpa/src/storage/metadata_mem.rs
@@ -114,7 +114,12 @@ impl storage::MetadataStorage for Storage {
 
     async fn poll_waiting(&self, tx: storage::Sender<bundle::Bundle>) -> storage::Result<()> {
         let mut entries = BTreeMap::new();
-        for bundle in self.entries.lock().iter().filter_map(|(_, bundle)| bundle.as_ref()) {
+        for bundle in self
+            .entries
+            .lock()
+            .iter()
+            .filter_map(|(_, bundle)| bundle.as_ref())
+        {
             if bundle.metadata.status == metadata::BundleStatus::Waiting {
                 entries.insert(bundle.metadata.read_only.received_at, bundle.clone());
             }
@@ -128,7 +133,11 @@ impl storage::MetadataStorage for Storage {
         Ok(())
     }
 
-    async fn poll_service_waiting(&self, source: Eid, tx: storage::Sender<bundle::Bundle>) -> storage::Result<()> {
+    async fn poll_service_waiting(
+        &self,
+        source: Eid,
+        tx: storage::Sender<bundle::Bundle>,
+    ) -> storage::Result<()> {
         let mut entries = BTreeMap::new();
         for bundle in self.entries.lock().iter().filter_map(|(_, v)| v.as_ref()) {
             if let metadata::BundleStatus::WaitingForService { service: s } =
--- a/sqlite-storage/src/storage.rs
+++ b/sqlite-storage/src/storage.rs
@@ -190,8 +190,8 @@ fn from_status(
             Some(source.to_string()),
         ),
         hardy_bpa::metadata::BundleStatus::Dispatching => (4, None, None, None),
-        hardy_bpa::metadata::BundleStatus::WaitingForService { source } => {
-            (5, None, None, Some(source.to_string()))
+        hardy_bpa::metadata::BundleStatus::WaitingForService { service } => {
+            (5, None, None, Some(service.to_string()))
         }
     }
 }


bpa/src/dispatcher/admin.rs
```diff
@@ -40,60 +40,70 @@ impl Dispatcher {
...
-                        service
-                            .on_status_notify(
-                                &report.bundle_id,
-                                &bundle.bundle.id.source,
-                                services::StatusNotify::Received,
-                                report.reason,
-                                assertion.0,
-                            )
-                            .await;
+                match self.service_registry.find(&report.bundle_id.source).await {
```
ricktaylor: I think we ought to do a Rib Lookup here, rather than just checking services.  If we are going to implement Active/Passive/None style keep/drop rules in the Rib, we need to check the Rib rather than short circuiting to the service_registry directly.

sylvain-pierrot: The RIB lookup is invoked just beforehand, at `process_bundle()` level. So concretely, the RIB has already decided where this bundle should go. For the Active/Passive services style, I assume the decision to invoke `administrative_bundle()` is only because the service is **Active**. That means, for the futur implementation **Active/Service services**, a service in passive state should be catch at `process_bundle()` level (in the pattern matching). 
Maybe I miss a step?

PR #452, opened 2026-03-27
refactor/bundle-transitions

sylvain-pierrot:
## Summary

Fix several bugs in the BPA and consolidate the bundle state machine.

**Bug fixes**
- Fragment reassembly: `next_offset` started from the wrong position, causing the completeness check to always fail
- In-memory storage: `get()` in duplicate check was silently promoting LRU entries; replaced with `peek()`
- Reaper: use `Store::get_metadata` instead of calling `metadata_storage` directly

**State machine consolidation**
- New `dispatcher/transitions.rs` is the single place where `BundleStatus` changes. Named transitions: `wait_for_route`, `wait_for_fragments`, `wait_for_service`, `tombstone`, `tombstone_with_report`
- `next_hop` moved from `ReadOnlyMetadata` to `BundleMetadata` with `#[serde(skip)]`. It is transient and must not be persisted
- Fixed a `_` wildcard in `dispatch.rs` that conflated `Deliver(None)` and `None (no route)`

### DIFF
--- a/bpa/src/dispatcher/forward.rs
+++ b/bpa/src/dispatcher/forward.rs
@@ -31,7 +31,7 @@ impl Dispatcher {
         // - If send fails or peer goes down, bundle returns to Waiting and may
         //   route to a different peer, so Egress will run again with fresh context
         // - BPSec blocks (BIB/BCB) should be added here, may be peer-specific
-        // - On Drop result: call drop_bundle() and return early
+        // - On Drop result: call tombstone_with_report() and return early
         let (bundle, data) = match self
             .filter_registry
             .exec(
@@ -46,7 +46,10 @@ impl Dispatcher {
         {
             filters::registry::ExecResult::Continue(_mutation, bundle, data) => (bundle, data),
             filters::registry::ExecResult::Drop(bundle, reason) => {
-                return self.drop_bundle(bundle, reason).await;
+                return match reason {
+                    Some(r) => self.tombstone_with_report(bundle, r).await,
+                    None => self.tombstone(bundle).await,
+                };
             }
         };
 
@@ -54,7 +57,7 @@ impl Dispatcher {
         match cla.forward(queue, cla_addr, data).await {
             Ok(cla::ForwardBundleResult::Sent) => {
                 self.report_bundle_forwarded(&bundle).await;
-                self.drop_bundle(bundle, None).await;
+                self.tombstone(bundle).await;
                 return;
             }
             Ok(cla::ForwardBundleResult::NoNeighbour) => {
--- a/bpa/src/dispatcher/forward.rs
+++ b/bpa/src/dispatcher/forward.rs
@@ -13,7 +13,7 @@ impl Dispatcher {
         // Get bundle data from store, now we know we need it!
         let Some(data) = self.load_data(&bundle).await else {
             // Bundle data was deleted sometime during processing
-            return self.delete_bundle(bundle).await;
+            return;
         };
 
         // Increment Hop Count, etc...
--- a/bpa/src/dispatcher/forward.rs
+++ b/bpa/src/dispatcher/forward.rs
@@ -45,8 +45,11 @@ impl Dispatcher {
             .trace_expect("Egress filter execution failed")
         {
             filters::registry::ExecResult::Continue(_mutation, bundle, data) => (bundle, data),
-            filters::registry::ExecResult::Drop(bundle, reason) => {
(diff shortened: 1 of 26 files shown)


ricktaylor (changes requested):

bpa/src/storage/adu_reassembly.rs
```diff
@@ -46,17 +46,14 @@ struct ReassemblyResult {
...
             source: bundle.bundle.id.source.clone(),
             timestamp: bundle.bundle.id.timestamp.clone(),
         };
 
         // See if we can collect all the fragments
-        let Some(results) = self.poll_fragments(bundle, &status).await else {
-            self.update_status(bundle, status).await;
-            return None;
-        };
+        let results = self.poll_fragments(bundle, &status).await?;
```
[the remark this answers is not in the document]
sylvain-pierrot: The caller will do it depending on:

```rust
/// Outcome of an ADU reassembly attempt.
///
/// Three distinct outcomes require three distinct caller actions — `Option`
/// would collapse `NotReady` and `Failed` into the same `None` arm, leading
/// the caller to incorrectly treat a failed (data-deleted) reassembly as
/// "wait for more siblings".
pub(crate) enum ReassemblyOutcome {
    /// Not all sibling fragments have arrived; fragment data is still in storage.
    /// Caller should transition the bundle to `AduFragment` and wait.
    NotReady,
    /// All fragments arrived and the ADU was successfully reassembled.
    Done(Arc<str>, Bytes),
    /// All fragments arrived but reassembly failed (corrupt/misaligned data).
    /// Fragment data has already been deleted; caller should drop the trigger bundle.
    Failed,
}
```

ricktaylor (changes requested):

ricktaylor (changes requested):

bpa/src/dispatcher/forward.rs
```diff
@@ -13,7 +13,7 @@ impl Dispatcher {
         // Get bundle data from store, now we know we need it!
         let Some(data) = self.load_data(&bundle).await else {
             // Bundle data was deleted sometime during processing
-            return;
+            return self.delete_bundle(bundle).await;
```
ricktaylor: Couldn't we just use `self.store.tombstone_metadata(&bundle.bundle.id).await` here?

ricktaylor (commented):
Very minor comments

bpa/src/dispatcher/mod.rs
```diff
@@ -152,7 +148,7 @@ impl Dispatcher {
                 if let Some(data) = dispatcher.load_data(&bundle).await {
                     dispatcher.ingest_bundle(bundle, data).await;
                 } else {
-                    dispatcher.drop_bundle(bundle, None).await;
+                    dispatcher.delete_bundle(bundle).await;
```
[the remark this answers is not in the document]
ricktaylor: I'm wrong, if load_data() returns None, then the bundle has already been cleaned up

bpa/src/dispatcher/forward.rs
ricktaylor: But perhaps we should report that it's gone, with a ReasonCode::DepletedStorage

bpa/src/dispatcher/transitions.rs
```diff
@@ -0,0 +1,94 @@
...
+use tracing::instrument;
+
+impl Dispatcher {
+    /// `* → Dispatching`: transition bundle into the dispatch pipeline.
+    ///
+    /// Used both as the Ingress-filter checkpoint (`New → Dispatching`) and to
+    /// re-dispatch bundles returning from wait states. Always persists full
+    /// metadata because the Ingress filter may have mutated metadata fields.
+    pub(super) async fn dispatching(&self, bundle: &mut Bundle) {
+        bundle.metadata.status = BundleStatus::Dispatching;
```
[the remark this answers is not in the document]
ricktaylor:
````rust
if bundle.metadata.status != BundleStatus::Dispatching { 
    bundle.metadata.status = BundleStatus::Dispatching; .
    .. 
}
````

bpa/src/dispatcher/forward.rs
sylvain-pierrot: I’ve reverted for now, do I do it?

sylvain-pierrot: Just `return` is enough as load_data invoke tombstone_metadata

sylvain-pierrot: ok done

PR #459, opened 2026-03-30
Tcpclv4 fuzz

ricktaylor:
- Restore TCPCLv4 fuzz testing (tcpclv4/fuzz/), adapted from the old tcpcl_old/fuzz/ harness (deleted in fe1d2e19) for the current crate architecture
  - Fix integer overflow bug in codec.rs extension parsing, found by the new fuzz harness
  - Add TCPCLv4 test coverage report with RFC 9174 compliance matrix
  - Uprev test plan documents to reflect current test infrastructure

  Fuzz harness

  Shared infrastructure (src/lib.rs) provides MockSink, MockBpa, and setup helpers for both targets. Session config tuned for fuzz throughput (2s contact timeout, no keepalive, no TLS).

  - passive — spawns a real TCPCLv4 listener on loopback, connects via TCP, writes fuzz-generated bytes as a malicious client
  - active — binds a fake server on loopback, triggers Cla::connect(), writes fuzz-generated bytes as a malicious server response

  Bug fix: subtraction overflow in extension parsing

  The fuzzer found a panic at codec.rs:673 — attempt to subtract with overflow in TransferSegmentMessage::decode. When a peer sends a transfer_extensions_byte_length that understates the actual extension data, ext_remaining -= 5 + item_length underflows.

  The same pattern existed in SessionInitMessage::decode.

  Fixed by using saturating_sub for ext_remaining and saturating_add for consumed in both extension parsing loops. No legitimate TCPCLv4 implementation would trigger this — it requires a malicious peer sending crafted extension metadata, which is why interop tests
  didn't catch it.

  Test documentation

  - New: tcpclv4/docs/test_coverage_report.md — RFC 9174 compliance matrix mapping all 10 LLRs to source files and interop evidence
  - Uprev: tcpclv4/docs/component_test_plan.md — added interop as current verification method, fixed UT-TCP-02 source file reference, scoped scaling tests to Full Activity, fixed execution commands
  - Uprev: tcpclv4/docs/fuzz_test_plan.md — updated target definitions to match actual implementation, added shared infrastructure section
  - Uprev: tcpclv4-server/docs/test_plan.md — noted SYS tests exercised by interop, scoped PKG-HELM-01 to Full Activity
  - Uprev: bpa/docs/cla_integration_test_plan.md — replaced non-existent cla_harness.rs reference with actual execution methods


tcpclv4/fuzz/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+/// The default listen address for fuzz targets.
+pub const FUZZ_ADDR: SocketAddr = SocketAddr::V6(std::net::SocketAddrV6::new(
+    std::net::Ipv6Addr::LOCALHOST,
+    4556,
+    0,
+    0,
+));
```
sylvain-pierrot: Using a fixed port here can lead to conflicts (already in use, parallel fuzzing, CI) :/
It would be more robust to bind to port 0 and propagate the assigned address to the fuzz targets instead of hardcoding 4556?

tcpclv4/src/codec.rs
```diff
@@ -156,14 +156,15 @@ impl SessionInitMessage {
...
             session_extensions.push(SessionInitExtension {
                 flags,
                 item_type,
                 item_length,
                 item_value: src_cloned.split_to(item_length as usize).into(),
             });
-            consumed += 5 + item_length as usize;
-            ext_remaining -= 5 + item_length as usize;
+            consumed = consumed.saturating_add(ext_size);
+            ext_remaining = ext_remaining.saturating_sub(ext_size);
```
sylvain-pierrot: Using saturating_* avoids overflow but changes the logic.
If ext_size > ext_remaining, this should be treated as invalid input instead?

tcpclv4/src/codec.rs
```diff
@@ -681,7 +683,7 @@ impl TransferSegmentMessage {
             return Ok(None);
         }
         // Skip the header bytes (message type, flags, transfer_id, extensions, data_length)
-        let _ = src.split_to(consumed + 8);
+        let _ = src.split_to(consumed.saturating_add(8));
```
sylvain-pierrot: Same concern here for saturation

tcpclv4/src/codec.rs
ricktaylor: Yeah - let me add a checked_sub and an error out rather that saturating_sub

ricktaylor: Changes:                         
                                                                                          
  - Added Error::InvalidExtensionLength variant
  - Both SESS_INIT and XFER_SEGMENT extension parsing: saturating_sub → checked_sub + Err(InvalidExtensionLength)                                                                                                                                                            
  - saturating_add on ext_size and consumed → plain + (can't overflow: 5 + u16::MAX and accumulating actually-parsed bytes)                                                                                                                                                  
  - Line 689: saturating_add(8) → + 8 (fixed header size)                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                             
  Malformed frames with ext_size > ext_remaining now return an error instead of silently clamping.

tcpclv4/src/codec.rs
ricktaylor: Fixed

tcpclv4/fuzz/src/lib.rs
ricktaylor: Changes:                  
                                                                                          
  - FUZZ_ADDR constant → fuzz_addr() function that checks FUZZ_LISTEN_ADDR env var, falling back to [::1]:4556                                                                                                                                                               
  - Both fuzz targets updated to call fuzz_addr() instead of using the constant                                                                                                                                                                                              
  - All changes are in tcpclv4/fuzz/ only — no production code touched

.gitignore
```diff
@@ -9,3 +9,6 @@ bpa-server/certs/
 # MkDocs
 /site/
 /.venv/
+
+# Fuzz corpus
+tcpclv4/fuzz/
```
sylvain-pierrot, line 14: Do we want to ignore the entire tcpclv4/fuzz/ directory here?

ricktaylor: No - I fixed this

sylvain-pierrot (approved):
Approved!

PR #476, opened 2026-04-09
ci: enable S3 storage in Docker image and skip workflows on non-code changes

sylvain-pierrot:
- Enable s3-storage feature in the Docker CI build for **hardy-bpa-server**
- Add path filters to `rust.yml` and `docker.yml` so they only run when Rust source, Cargo files, Dockerfiles, or protobuf definitions change


ricktaylor (approved):
LGTM

Issue #479, opened 2026-04-09
register_service API does not support multi-endpoint services


ricktaylor:
A single service implementation registered at multiple endpoints (e.g., echo at both ipn:1.7 and ipn:1.2047) receives a separate on_register call with a separate Sink for each endpoint. Dropping any sink triggers unregistration of that endpoint via Sink::Drop.       
   
  This forces every multi-endpoint service to either:                                                                                                                                                                                                                        
  - Store all sinks (even though it only needs one to send), or
  - Create separate instances per endpoint (current workaround in bpa-server/src/services.rs)                                                                                                                                                                                
                                                                                             
  The API should support registering a single service at multiple endpoints with one on_register / one sink. Options:                                                                                                                                                        
                                                                                                                                                                                                                                                                             
  - register_service(service_ids: &[Service], service: Arc<dyn Service>) -> Result<Vec<Eid>> — single call, one sink, multiple routes                                                                                                                                        
  - Detect same Arc<dyn Service> already registered — add route, skip on_register, share existing sink                                                                                                                                                                       

Others:                                                                                                                                                                                      
- Add `fn register_dynamic_service`.
                                                                                       
Current workaround: bpa-server creates a new EchoService instance per service ID.                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             Related: service registration redesign #478

ricktaylor:
Closing as by-design. The current model is consistent with the Sink pattern used throughout Hardy (CLAs, routing agents, services all follow the same trait + sink contract):

- Each service ID is a separate endpoint with its own route in the RIB, its own EID, and its own sink. Dropping a sink unregisters that specific endpoint — this is correct granularity.
- Multi-endpoint services that share state wrap it in an Arc and create a lightweight instance per endpoint. This is idiomatic Rust and cheap (EchoService::new() is trivial).
- The proposed single-sink-multiple-endpoint API would couple endpoint lifetimes together, preventing independent deregistration. The current model is more flexible.

The register_dynamic_service API was added in #519. The echo "workaround" (separate instance per endpoint) is the intended pattern, not a limitation.

PR #480, opened 2026-04-09
Fix HDTN interop tests

ricktaylor:
HDTN interop test fixes                                                                                                                                                                                                                                                    
                                         
  - Fix echo service unregistration bug — EchoService.register() now creates a separate instance per endpoint, preventing Sink::Drop from unregistering the second service (workaround for #479)                                                                             
  - Fix connection refused on test 1 — add 2s delay after ss detects HDTN listening, as the socket is in LISTEN state before the accept loop is ready                                                                                                                        
  - Fix benchmark build — remove stale dynamic-plugins feature flags (reverted in a previous PR). Clear binary paths on build failure to prevent stale binaries producing duplicate results                                                                                  
  - Add ud3tn to benchmark notes — update CLA protocol description in results footer


bpa-server/src/services.rs
```diff
@@ -10,15 +10,13 @@ pub async fn register(
...
-                {
-                    Ok(eid) => info!("Echo service registered at {eid}"),
-                    Err(e) => error!("Failed to register echo service at {service}: {e}"),
+            match hardy_echo_service::EchoService::register(bpa, services).await {
+                Ok(eids) => {
+                    for eid in eids {
+                        info!("Echo service registered at {eid}");
+                    }
                 }
+                Err(e) => error!("Failed to register echo service: {e}"),
```
sylvain-pierrot, line 19: This changes the behavior: previously each service was registered independently, while now a failure aborts the whole batch

sylvain-pierrot: Because of `.await?` in `EchoService::register()`

echo-service/src/lib.rs
```diff
@@ -21,19 +21,16 @@ impl EchoService {
...
         bpa: &dyn BpaRegistration,
         services: &[hardy_bpv7::eid::Service],
     ) -> Result<Vec<hardy_bpv7::eid::Eid>, hardy_bpa::services::Error> {
         let mut eids = Vec::with_capacity(services.len());
         for service in services {
-            let eid = bpa
-                .register_service(Some(service.clone()), self.clone())
-                .await?;
+            let echo = Arc::new(Self::new());
+            let eid = bpa.register_service(Some(service.clone()), echo).await?;
```
sylvain-pierrot, line 33: Just curious, why regirster_service takes an Option as service?

bpa-server/src/services.rs
ricktaylor: Yeah - this changes the behaviour.  Previous each subsequent register unregistered the previous echo service, which was an issue.

This method now registers a unique echo service instance for each service number

echo-service/src/lib.rs
ricktaylor: No idea - let me check

ricktaylor: None = let the BPA auto-assign a service number/name. Useful for services that don't need a well-known endpoint (e.g. netops tool service, dynamically registered gRPC services). Some(id) = register at a specific service ID (e.g. echo at service 7).

sylvain-pierrot: Ok makes sense, that's what I thought. What do you think about introducing `fn register_dynamic_service` instead (later)?

bpa-server/src/services.rs
sylvain-pierrot: Alright 🙂

echo-service/src/lib.rs
ricktaylor: Yeah - later definitely - there is a whole bunch of design improvements to do with services.  I think maybe &[Service] and allow empty...  maybe? #479

sylvain-pierrot: We could capture that in this issue as well?

ricktaylor: Yes please!

ricktaylor:
Just added a load of other interop test cleanup into one PR

sylvain-pierrot (approved):
LGTM

PR #492, opened 2026-04-14
refactor(bpa): typestate registries and full builder pattern

sylvain-pierrot:
## Summary

- [**BPA**]: Migrate to a full builder pattern. Previously, all services, filters, and CLAs had to be registered manually on a live BPA, even though they're all known at startup from the config. Now the builder collects them first, and `build()` creates the BPA with everything already in place. Subsystem constructors are simple functions that take individual parameters and return their component, with all wiring visible in [bpa-server/mod.rs](https://github.com/ricktaylor/hardy/blob/refactor/bpa-builder-and-registries/bpa-server/src/bpa/mod.rs).
- [**BPA Server**]: Add a `policies` field in BPA Server configuration. Policies are defined once by name and shared across CLAs by reference.
- [**BPA**]: Fixed several ownership issues : registries used to store `Arc` references to the Rib, Store, and Dispatcher as struct fields. Now these are passed as parameters instead. One circular dependency remains (Dispatcher ↔ CLA registry), mitigated with deferred init instead of storing an `Arc` in the struct.
- [**BPA**]: Add dynamic service/application registration with auto-assigned service IDs.

### Builder pattern

Before, you had to create a BPA, then call `register_*()` for every CLA, service, filter... even though you already know all of them from the config. There was no separation between "setting up" and "running".

Now it's two steps:

- **Build** : you declare everything through the builder (`builder.cla(...)`, `builder.service(...)`, `builder.filter(...)`, etc.). Nothing starts yet. Then `build()` constructs all registries in parallel and gives you a ready-to-go BPA.
- **Run** : the BPA is live. Only things that actually arrive at runtime (gRPC client registering a service, plugin loading a CLA) go through `register_*()` on `BpaRegistration`.

### Named policies

Before, each CLA had its own inline policy config. If two CLAs needed the same policy, you'd duplicate the config and get two separate instances.

Now, policies are defined once by name in the top-level config and CLAs reference them by name. Same name = same `Arc`, shared instance. If a CLA references an unknown policy name, startup fails immediately.

### Ownership fixes

The CLA and Service registries used to store `Arc` references to the Rib, Store, and Dispatcher as struct fields. Now these are passed as parameters to the methods that need them instead of being held permanently. This makes dependencies explicit and reduces coupling.

One circular dependency remains between Dispatcher and CLA registry : Dispatcher needs CLA registry for forwarding, CLA registry needs Dispatcher for peer queue consumers. Currently mitigated with deferred init (`Once`). The root cause is that Dispatcher does too much : see `dispatcher_decomposition.md` for the longer-term fix.

### Dynamic registration

`register_service()` and `register_application()` now require an explicit service ID : no more `Option<Service>`.

For callers that don't care about a specific ID (like `bp ping` or BIBE when no service is configured), there's `register_dynamic_service()` and `register_dynamic_application()` which auto-assign an IPN service number.


ricktaylor:
bpa/src/builder.rs — cla() method                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  The address_type parameter has moved from register_cla() (where the CLA declares its own capability) to the builder (where bpa-server config controls it). This is a regression.
                                                                                                                                                                                                                                                                             
  The CLA is the authority on what address types it can handle — TCPCLv4 knows it handles Tcp, a future UDPCLv2 knows it handles Udp. This is intrinsic to the CLA implementation, not an administrative decision.                                                           
                                                                                                                                                                                                                                                                             
  With this change:                                                                                                                                                                                                                                                          
  - bpa-server config decides what address types a CLA supports (wrong — the CLA knows its own capabilities)
  - gRPC-registered CLAs can't declare their address types (the server would have to guess)                                                                                                                                                                                  
  - The clas::new() function in bpa-server returns the address type, duplicating knowledge that belongs in the CLA
                                                                                                                                                                                                                                                                             
  Suggest keeping address_type on register_cla() where the CLA passes it in, or adding a method to the Cla trait (fn address_type() -> Option<ClaAddressType>) so the BPA queries the CLA for its capability at registration time.

README.md
ricktaylor, line 1: I'm going to make this change direct on main for the release

ricktaylor (changes requested):
I left a comment on the PR - there is a regression with ClaAddressType registration that needs fixing

bpa-server/src/bpa/clas.rs
```diff
@@ -34,64 +33,44 @@ pub enum ClaConfig {
...
+#[allow(unused_variables)]
+pub fn new(name: &str, cla_type: &ClaType) -> anyhow::Result<Option<NewClaResult>> {
+    match cla_type {
+        #[cfg(feature = "tcpclv4")]
+        ClaType::TcpClv4(config) => {
+            let cla = Arc::new(
+                hardy_tcpclv4::Cla::new(config)
+                    .map_err(|e| anyhow::anyhow!("Failed to create CLA '{name}': {e}"))?,
+            );
+            Ok(Some((cla, Some(hardy_bpa::cla::ClaAddressType::Tcp))))
```
ricktaylor: This is the issue:  The Cla should be the the source of truth for the ClaAddressType

sylvain-pierrot: Thanks!

sylvain-pierrot:
I’ve resolved your issue. Let me know your opinion. Basically, I’ve moved `address_type` to the Cla trait, as it’s an internal matter

bpa/src/cla/mod.rs
```diff
@@ -194,6 +194,11 @@ pub trait Cla: Send + Sync {
     /// and release resources. After this returns, the Sink is no longer functional.
     async fn on_unregister(&self);
 
+    /// Returns the address type this CLA handles, if any.
```
ricktaylor, line 197: YEs - nice and clean!

bpa/src/cla/registry.rs
```diff
@@ -155,47 +209,39 @@ impl Registry {
...
                 policy: policy
                     .unwrap_or_else(|| Arc::new(policy::null_policy::EgressPolicy::new())),
-            }))
-            .clone()
-        };
+            }));
+        }
 
-        metrics::gauge!("bpa.cla.registered").increment(1.0);
+        let cla = self.clas.lock().get(&name).cloned().unwrap();
```
ricktaylor: Surely we can return this from the previous lock() without locking again?

bpa/src/cla/registry.rs
```diff
@@ -155,47 +209,39 @@ impl Registry {
...
 
-        // Register that the CLA is a handler for the address type
         if let Some(address_type) = address_type {
```
ricktaylor, line 240: Let's call address_type() here - no need to call it earier

bpa/src/cla/registry.rs
```diff
@@ -222,25 +271,20 @@ impl Registry {
...
-        if let Some(address_type) = &cla.address_type {
-            self.rib.remove_address_type(address_type);
+        if let Some(address_type) = cla.cla.address_type() {
+            self.rib.remove_address_type(&address_type);
         }
 
         let peers = core::mem::take(&mut *cla.peers.lock());
 
         for (_, (node_ids, peer_id)) in peers {
-            // Remove RIB entries for all EIDs associated with this address
```
ricktaylor, line 232: Can we keep the comments?  They were useful

bpa/src/node_ids.rs
```diff
@@ -52,31 +52,27 @@ impl NodeIds {
...
-                    service_number,
-                },
-                Some((a, n)),
-                _,
-            ) => allocator_id == a && node_number == n && service_number == &0,
-            (Eid::Dtn { node_name, demux }, _, Some(n)) => node_name == n && demux.is_empty(),
-            _ => false,
+    /// Resolve a service identifier to a full EID using this node's identity.
+    pub fn resolve_eid(&self, service_id: &hardy_bpv7::eid::Service) -> Eid {
+        match service_id {
```
ricktaylor, line 63: You can use ` TryFrom<(NodeId, Service)> for Eid ` here ;)

ricktaylor (changes requested):
Minor comments, but overall this works

bpa/src/cla/registry.rs
sylvain-pierrot: Yeah sure!

bpa/src/cla/registry.rs
sylvain-pierrot: In this method there is a little confusion regarding the naming, in parameter we have `cla` -> `Arc<dyn Cla>`, and here we have `cla` variable -> `Arc<Cla>` (not the trait).

I can call it later but the value moves just beforehand, so need to clone() it. I’d rather not clone it, right?

bpa/src/cla/registry.rs
sylvain-pierrot: I’ll bring back the comment

bpa/src/node_ids.rs
sylvain-pierrot: I already need to retrieve the node ID based on the service ID, I think tryfrom doesn’t match this use case :/

bpa/src/cla/registry.rs
sylvain-pierrot: Done

bpa/src/node_ids.rs
ricktaylor: Hmmm... the expect() call worries me.  Can we return an error, or is this prechecked?

ricktaylor (commented):
Quick comment

bpa/src/node_ids.rs
sylvain-pierrot: To be honest, me too … and I have a generic question regarding that. At this step, the BPA should own a nodeid, and because of `Option` in `struct NodeIds` we need `.expect`. 
So a BPA can choose either dtn or ipn, or both are mandatory, or at least one is mandatory?
I’m thinking about removing these panics...

sylvain-pierrot: Or otherwise, I just return an error like you said

ricktaylor: Only one is required - either ipn or dtn...  But only 1 of each.

sylvain-pierrot: OK, I return an error then

sylvain-pierrot: Done!

ricktaylor (approved):
Looks great!

PR #500, opened 2026-04-16
Tcpclv4 metrics

ricktaylor:
Add metrics to tcpclv4


sylvain-pierrot:
Overall very good!

One concern: I think the `session.bytes.sent/received` naming is a bit stronger than what the code currently measures. In the codec, the bytes encoded/decoded are counted at the framing layer, not actual socket I/O

ricktaylor:
Although it's counted at the framing layer, it's the raw TCP bytes, not the TCPCL payload.  Okay - no IP/TCP header information, but that's usual

sylvain-pierrot (approved):
LGTM!

PR #501, opened 2026-04-17
refactor(bpa): explicit gRPC lifecycle

sylvain-pierrot:
### gRPC server lifecycle

Replaces `init()` with a `GrpcServer` struct that separates construction from execution. Errors are now propagated instead of panicking, and the caller controls task spawning. See https://github.com/ricktaylor/hardy/blob/934e3e18e3f05d2e0b5779b054dfdf65641115d4/bpa-server/src/main.rs#L56


bpa-server/src/bpa/mod.rs
ricktaylor, line 1: Given we just have this short mod.rs file - why not move it up one level, and call it bpa.rs avoiding the whole directory?  Might be cleaner?

ricktaylor (commented):
Single comment - open to feedback

bpa-server/src/bpa/mod.rs
```diff
@@ -1,26 +1,23 @@
...
 
 pub(crate) mod clas;
-pub(crate) mod grpc;
 pub(crate) mod policy;
 pub(crate) mod static_routes;
 pub(crate) mod storage;
 
-pub(crate) async fn run(
+/// Build a BPA from the given configuration.
+pub(crate) async fn build(
```
ricktaylor: No need for `pub(crate)` can just be `pub` we are a binary

proto/src/server/mod.rs
```diff
@@ -42,62 +42,75 @@ impl Default for Config {
...
+        })
     }
 
-    let (health_reporter, health_service) = tonic_health::server::health_reporter();
-    // Start serving
-    let addr = config.address;
-    let cancel_token = tasks.cancel_token().clone();
-    hardy_async::spawn!(tasks, "grpc_server", async move {
+    /// Serve until cancelled, then shut down session tasks.
+    pub async fn serve(
```
ricktaylor, line 100: Yes - I like this!

ricktaylor (changes requested):
Minor comment on pub(crate) when we are a binary

bpa-server/src/bpa/mod.rs
sylvain-pierrot: Totally agree! I haven’t created a `bpa.rs` though. Instead I’ve wrapped into a `config` folder, because it’s all config related. Check out the new structure 🙂

bpa-server/src/bpa/mod.rs
sylvain-pierrot: Indeed. I’ve done the change

ricktaylor (approved):
LGTM

PR #503, opened 2026-04-17
refactor(bpa): redesign recovery with typestate pattern and fast startup 

sylvain-pierrot:
## Description

Extracts recovery into a dedicated `bpa/src/recover/` module, redesigns crash recovery to be fast by default, and introduces safe operation ordering that guarantees metadata always points to valid data.

### Recovery module

Recovery lives in `bpa/src/recover/` and uses a typestate pattern. The `Recovery` struct has three states, each in its own file:

```
Recovery<Idle>       idle.rs       - entry point, selects recovery path
Recovery<Marked>     marked.rs     - all metadata marked as unconfirmed
Recovery<Confirmed>  confirmed.rs  - all bundle data walked and reconciled
```

Each transition consumes the previous state at compile time. Phases cannot be skipped or reordered. The struct borrows `Store` and `Dispatcher` with zero allocation.

Usage example:

```rust
crate::recover::Recovery::new(&store, &dispatcher)                                                                                              
      .mark().await        // Idle -> Marked
      .reconcile().await   // Marked -> Confirmed                                                                                                 
      .purge().await;      // Confirmed -> done 
```

### Safe operation ordering

All cross-store operations now order writes so that a crash always results in an orphan data file (harmless) rather than metadata pointing to missing data:

| Operation       | Order                                     | Crash outcome    |
| --------------- | ----------------------------------------- | ---------------- |
| Store           | save data -> insert metadata              | Orphan data file |
| Delete          | tombstone metadata -> delete data         | Orphan data file |
| Filter mutation | save new -> update metadata -> delete old | Orphan data file |

In all cases, metadata always points to valid data. Orphan files are detected and recovered or cleaned up. This eliminates false `DepletedStorage` reports on crash recovery.

## Bug fixes

**SQLite `confirm_exists` panics on tombstoned rows:** The SELECT did not filter `bundle IS NOT NULL`. A tombstoned row caused `row.get::<_, Vec<u8>>` to panic. Fixed by adding `AND bundle IS NOT NULL` and wrapping the DELETE + SELECT in an explicit transaction (also fixes the non-atomic issue).

**Orphaned data for tombstoned bundles:** `insert_metadata` returned `false` for tombstoned bundle IDs but the return value was discarded — the data file was never deleted and rediscovered on every restart. Fixed by checking the return value and deleting the orphan.

**Non-fsync path skips .tmp rename:** When `fsync: false`, files were written directly to their final name. A crash left partial files indistinguishable from valid data. Fixed by always using the `.tmp` rename protocol — the flag now only controls whether sync calls are made.


bpa/src/dispatcher/mod.rs
```diff
@@ -162,7 +161,7 @@ impl Dispatcher {
         });
     }
 
-    fn key_provider(
+    pub(crate) fn key_provider(
```
ricktaylor, line 166: If Dispatcher is pub(crate) this can just be pub

ricktaylor: However, pub(super) might be sufficient to allow other Dispatcher functions to access without exposing globally

bpa/src/dispatcher/report.rs
```diff
@@ -5,7 +5,7 @@ use hardy_bpv7::status_report::{
 
 impl Dispatcher {
     #[cfg_attr(feature = "instrument", instrument(skip(self, bundle),fields(bundle.id = %bundle.bundle.id)))]
-    pub(super) async fn report_bundle_reception(
+    pub(crate) async fn report_bundle_reception(
```
ricktaylor, line 8: See above

bpa/src/dispatcher/dispatch.rs
```diff
@@ -203,7 +203,7 @@ impl Dispatcher {
     ///
     /// Because this returns before the Ingress filter completes, bundles remain
     /// in `New` status until `ingest_bundle_inner()` checkpoints to `Dispatching`.
-    pub(super) async fn ingest_bundle(self: &Arc<Self>, bundle: bundle::Bundle, data: Bytes) {
+    pub(crate) async fn ingest_bundle(self: &Arc<Self>, bundle: bundle::Bundle, data: Bytes) {
```
ricktaylor, line 206: pub(super) is used to make this function available to other Dispatcher functions outside this file, but private to the Dispatcher module.  If it's needed across the crate, then just `pub` is sufficient, as Dispatcher is `pub(crate)`

bpa/src/dispatcher/dispatch.rs
```diff
@@ -295,7 +295,7 @@ impl Dispatcher {
     }
 
     /// Queue a bundle for dispatch processing
-    pub(super) async fn dispatch_bundle(&self, mut bundle: bundle::Bundle) {
+    pub(crate) async fn dispatch_bundle(&self, mut bundle: bundle::Bundle) {
```
ricktaylor, line 299: See above

ricktaylor:
This should close because both commits were rebased onto #505 (refactor/dispatcher), which has already been closed for its design conflict with the queue/streaming architecture on main — this PR inherits the same conflict. The three-phase recovery typestate idea is worth revisiting once the queue architecture is implemented; it's noted in the audit record so it isn't lost. Branch will be deleted with the close.

ricktaylor (changes requested):

PR #507, opened 2026-04-21
refactor(fragmentation): progressive reassembly with direct write-at-offset and coverage tracking 

sylvain-pierrot:
## Description

Replaces the collect-then-stitch fragment reassembly with a **progressive approach**: each fragment writes its payload **directly to its final ADU position** in storage. Coverage is tracked in a single metadata row. When all bytes are covered, the bundle is finalized.

### RFC 9171 compliance

Fragment reassembly follows RFC 9171 Section 5.9:

- **Overlapping fragments**: the RFC allows fragments from different fragmentation episodes to produce overlapping payload subsets. The `Coverage` interval set handles this correctly by merging overlapping byte ranges, unlike the previous `HashMap<offset, fragment>` which silently overwrote.
- **Fragment 0 extension blocks**: per RFC 9171, fragment 0 carries all extension blocks from the original bundle (including BPSec Block Integrity/Confidentiality Blocks). Its full wire data is stored in the reassembly metadata entry for primary block reconstruction at finalization.
- **Single fragmentation level**: the RFC specifies only one level of fragmentation (as in IP). The coverage model tracks byte ranges against `total_adu_length` from the fragment's primary block, independent of how many fragmentation episodes produced the fragments.

### Performance

Backend storage operations for N fragments:

| | Previous | Progressive |
|---|---|---|
| Data writes | N+1 | N+1 (1 create + N write_at) |
| Metadata writes | 2N | N+1 (N upsert + 1 set_name) |
| Metadata reads | 1 table scan | 0 |
| Data reads | N (load all to stitch) | 1 (finalize) |
| Data deletes | N (fragment cleanup) | 1 (ADU object) |
| Metadata deletes | N (tombstones) | 1 (reassembly row) |
| **Total I/O** | **4N+2** | **N+5** |

For 100 fragments: 402 → 105 operations. For 1000 fragments: 4002 → 1005.

The key wins beyond the count: no table scan on every fragment arrival, no in-memory stitch step (the ADU is already assembled on storage), and no per-fragment cleanup phase.

### Multi-replica deployments

The progressive approach works across BPA replicas sharing a database and object store (e.g. Postgres + S3):

- Each replica writes its fragment's payload to S3 via `write_at`
- Each replica updates the same Postgres reassembly row atomically via `upsert_reassembly`
- The replica that completes the coverage triggers finalization
- No coordination needed beyond the atomic metadata upsert

This is the same coordination model as the previous implementation (shared database as the synchronization point), but with fewer round-trips per fragment.

### Behavioral differences

- **No reaper monitoring for pending fragments**: incomplete reassemblies won't be expired automatically. Future work: TTL sweep for stale reassembly entries.
- **No per-fragment status metrics**: the `bpa.bundle.status` gauge no longer shows `AduFragment` state. Future work: dedicated `bpa.reassembly.pending` gauge.
- **Overlap handling**: correctly merges overlapping byte ranges per RFC 9171 (previous implementation silently overwrote).

### Tests

26 tests total:
- **Coverage interval set** (14 tests): insert, merge, overlap, gaps, zero-length, adjacent, reverse order, completeness
- **Progressive reassembly** (12 tests): pending/complete lifecycle, reverse order, duplicates, non-fragment rejection, 5 and 10 fragment reassembly, payload byte-level verification, independent concurrent bundles, shuffled arrival order, cleanup after completion


bpa/src/dispatcher/dispatch.rs
```diff
@@ -371,8 +371,16 @@ impl Dispatcher {
             Some(rib::FindResult::Deliver(Some(service))) => {
                 // Check for reassembly
                 if bundle.bundle.id.fragment_info.is_some() {
-                    // Reassemble the bundle before delivery
-                    self.reassemble(bundle).await
+                    match crate::fragmentation::Reassembler::new(&self.store, self.key_provider())
```
ricktaylor: Hmmm....  this introduces a heap allocation per reassembly...  that's unecessarilly expensive

ricktaylor (changes requested):
I like it - but can we not new (heap allcocate) on each fragmented bundle?

bpa/src/dispatcher/dispatch.rs
sylvain-pierrot: Indeed, performance issue here!!

sylvain-pierrot: Now the same Reassembler is used for each reassembly!

ricktaylor: Actually - I was wrong - it was stack allocated (no Box<>) so it was just a store.clone() - sorry I think the new() was actually cleaner!

Can we go back to how you had it

bpa/src/storage/mod.rs
```diff
@@ -309,8 +308,8 @@ struct BundleCache {
 
 // Storage helper
 pub(crate) struct Store {
-    tasks: hardy_async::TaskPool,
-    metadata_storage: Arc<dyn storage::MetadataStorage>,
+    pub(crate) tasks: hardy_async::TaskPool,
```
ricktaylor: Can just be pub - the struct is pub(crate) so the members cannot be > pub(crate)

ricktaylor (changes requested):
Sorry - I was wrong on my previous review - and a small comment

ricktaylor:
Okay - let me continue to review, but immediate comments:

1) I am a little concerned by the lack of reaper expiry - could we do something such as writing a 'best so far' metadata entry for the bundle as we reassemble it, therefore the reaper should pick it up?

2) I think we can add the new gauge as part of this change.

3) Does this impact the BundleStatus enum by removing ReassemblyPending?  If so - I think we should make the change in the backends as well

ricktaylor:
I'm not so sure this is worth the update.  Given fragmentation is effectively deprecated to be replaced with BIBE, this introduces a whole level of complexity for storage trait implementors that we should avoid.

I haveno problem with progressively rewriting bundles as fragments arrive, but I would rather do it using ::load() and store(), or replace(), with some kind of updated PrimaryBlock as required, rather than these more radical changes

ricktaylor:
This should close per the existing review consensus: progressive reassembly with write-at-offset conflicts with streaming_pipeline_design.md §2.4 — bundle data I/O is fully sequential and never modified in place, across all four storage backends this PR touches. Reassembly hardening remains wanted, but it has to be built on the sequential-spool model. Branch will be deleted with the close.

PR #509, opened 2026-04-23
refactor(filters): extract validity checks into filter + CBOR pre-check

sylvain-pierrot:


bpa/src/builder.rs
```diff
@@ -217,26 +216,36 @@ impl BpaBuilder {
...
         {
             use crate::filters::rfc9171::Rfc9171ValidityFilter;
 
-            filter_registry
+            filter_engine
                 .register(
                     Hook::Ingress,
                     "rfc9171-validity",
-                    &[],
+                    &["bundle-validity"],
```
ricktaylor: Does this have to be afterwards?  Could it be concurrent?

bpa/src/dispatcher/dispatch.rs
```diff
@@ -234,31 +202,9 @@ impl Dispatcher {
     /// filter execution details.
     #[cfg_attr(feature = "instrument", instrument(skip_all,fields(bundle.id = %bundle.bundle.id)))]
     pub(super) async fn ingest_bundle_inner(&self, mut bundle: bundle::Bundle, mut data: Bytes) {
-        if let Some(u) = bundle.bundle.flags.unrecognised {
```
ricktaylor, line 237: I think leave the unrecognised flags debug! here, because the filter can be used for originate as well

bpa/src/builder.rs
```diff
@@ -217,26 +216,36 @@ impl BpaBuilder {
...
     }
 }
 
 impl Default for BpaBuilder {
     fn default() -> Self {
-        let filter_registry = Arc::new(FilterRegistry::new());
+        let filter_engine = Arc::new(FilterEngine::new());
+
+        // Auto-register bundle validity filter (lifetime, hop-count)
+        filter_engine
```
ricktaylor, line 232: I think we can register for originate as well - we should be discarding invalid bundles from Services as well

ricktaylor (changes requested):
A few comments - but I like it!

bpa/src/builder.rs
sylvain-pierrot: Good catch! I’ve added it

bpa/src/builder.rs
sylvain-pierrot: Indeed, I’ve removed the dep!

bpa/src/dispatcher/dispatch.rs
sylvain-pierrot: I’ve moved it in the bundle filter validity to gather all validations. I feel like it’s a filter concern

sylvain-pierrot: And I’ve added the filter on originate as well

ricktaylor: But it's not a concern in the Originate path - so if we leave it here, we won't get spurious debug! in the Originate?

bpa/src/builder.rs
```diff
@@ -232,9 +233,18 @@ impl Default for BpaBuilder {
                 Hook::Ingress,
                 "bundle-validity",
                 &[],
-                Filter::Read(Arc::new(crate::filters::validity::BundleValidityFilter)),
+                Filter::Read(Arc::new(BundleValidityFilter)),
```
ricktaylor: Do we need a new Filter instance, or can we clone() the Arc?

ricktaylor (commented):
Left a question

bpa/src/dispatcher/dispatch.rs
sylvain-pierrot: Oh I see your point. You mean it’s only for Ingress?

sylvain-pierrot: Originate is for locally-generated bundles, if the `BundleValidityFilter` trigger a drop, that’s a programming bug right?

ricktaylor: Not necessarilly - we could have a thrid-party gRPC-based service, so filtering is applicable

bpa/src/builder.rs
sylvain-pierrot: Yes!! I’ve made the change to clone an arc

bpa/src/dispatcher/dispatch.rs
sylvain-pierrot: I suggest not splitting logic across two places. Either we can keep it in `BundleValidityFilter` or create a dedicated filter for observability?

ricktaylor: No - the basic checks are the same - I think it's worth having it reused

ricktaylor: Let's leave it in BundleValidityFilter until we know it's a problem

ricktaylor (approved):
LGTM!

PR #520, opened 2026-04-29
Design for a queue-based approach

ricktaylor:
Just a markdown document of a way forwards for consistency and performance...


ricktaylor:
No longer needed

PR #524, opened 2026-04-30
Big rework and tidy-up of ingress

### LINKED ISSUE #522: BPA returns bundle validation errors to CLA instead of handling internally
## Problem

When a CLA dispatches a bundle to the BPA via `Sink::dispatch()`, the BPA returns errors for invalid bundle *content* (malformed CBOR, parse failures) back to the CLA as `cla::Error` / `google.rpc.Status`. The CLA cannot do anything about a malformed bundle - it just received bytes from the wire and faithfully forwarded them. Returning an error forces the CLA to handle a condition it has no ability to fix, and conflates "your request was malformed" with "the bundle you carried was garbage."

The code is already aware of this - there's a TODO at `bpa/src/dispatcher/dispatch.rs:26`:

```rust
// TODO: Really should not return errors when the bundle content is garbage -
// it's not the CLAs responsibility to fix it!
```

Notably, the code already handles the *partially-valid* case correctly: when `RewrittenBundle::parse()` returns `Invalid { bundle, reason, error }`, the bundle is dropped internally with metrics and a status report (lines 99-124, 153-156). The problem is only with the two early-return error paths.

## Affected Code Paths

1. CBOR pre-check failure (`dispatch.rs:36-39`)

```rust
if let Err(e) = crate::cbor::precheck(&data) {
    metrics::counter!("bpa.bundle.received.dropped").increment(1);
    return Err(e.into());  // ← returned to CLA
}
```
Errors: empty payload, possible BPv6 bundle, non-CBOR-array first byte.

2. Bundle parse failure (dispatch.rs:42-46)
```rust
match hardy_bpv7::bundle::RewrittenBundle::parse(&data, self.key_provider()) {
    Err(e) => {
        metrics::counter!("bpa.bundle.received.dropped").increment(1);
        return Err(e.into());  // ← returned to CLA
    }
    ...
}
```
Errors: any hardy_bpv7::Error from malformed bundle structure.

3. gRPC error propagation (proto/src/server/cla.rs:53 and :183)
Both paths convert the cla::Error into a tonic::Status (line 53) or a google.rpc.Status proto message (line 183), sending it back to the CLA over the gRPC stream.

## Correct Behavior

When the bundle content is garbage, the BPA should:
1. Increment bpa.bundle.received.dropped (already done)
2. Log the error at debug! level (for operator diagnostics)
3. Return Ok(()) to the CLA

No status report can be generated for unparseable bundles since the source EID is unknown. The Invalid variant (lines 99-124) already handles the case where enough of the bundle was parsed to extract a source EID.

### bpa/src/dispatcher/dispatch.rs

Change both early-return error paths to log and return Ok(()):

- Lines 36-39: CBOR precheck -- log at debug, return Ok(())
- Lines 42-46: Parse failure -- log at debug, return Ok(())

### bpa/src/cla/mod.rs

Consider removing InvalidBundle(#[from] hardy_bpv7::Error) from cla::Error (line 24-25). After this fix, bundle validation errors should never flow through the CLA error type. If other callers need this variant it can stay, but it should not be reachable from the dispatch path.

### Return type of receive_bundle()

Once bundle validation errors are handled internally, receive_bundle() may not need to return cla::Result<()> at all -- it could return Result<(), cla::Error> with only the Disconnected and Internal variants possible, or even be infallible from the CLA's perspective. This is a design choice worth considering but not strictly required.

### No proto changes needed

The DispatchBundleResponse is already an empty message. The gRPC error path (Status message) remains available for genuine infrastructure errors (disconnection, internal failures). Only the content of what triggers an error response changes.

ricktaylor:
## Summary

- **CLA error handling**: `receive_bundle()` no longer returns bundle validation errors to the CLA — invalid bundles are logged, counted, and dropped internally since the CLA cannot fix them
- **Shared ingress pipeline**: Extracted `process_received_bundle()` so CLA ingress and fragment reassembly share the same parse/validate/store/report path
- **Reassembly correctness**: Reassembled bundles now go through `RewrittenBundle::parse()` (Full mode) instead of `ParsedBundle::parse()` (Preserve mode), getting block removal, canonicalization, BPSec, status reports, and Ingress filtering
- **Origination filter bypass**: Originated bundles store with `Dispatching` status and skip the Ingress filter — previously they stored as `New` and incorrectly ran Ingress on restart
- **Status report filter bypass**: Same fix as origination — store with `Dispatching`, skip Ingress filter
- **Removed unnecessary spawn**: `ingress_bundle()` runs inline in the caller's context; backpressure comes from the RpcProxy handler pool
- **Restart recovery simplification**: `restart_bundle()` uses `ParsedBundle` for integrity checks (not full reprocessing), preserves original metadata for known bundles, and sends orphans through the full receive pipeline via `process_received_bundle()`
- **Restart crash safety**: Non-canonical data replaced in-place via `replace_data()` instead of save-then-delete with a data-loss window
- **Code organisation**: Ingress functions moved from `dispatch.rs` to `ingress.rs`; renamed `ingest_bundle` → `ingress_bundle` for consistency
- **Removed `InvalidBundle` from `cla::Error`**: Bundle validation errors no longer flow through the CLA error type
- **Explicit `BundleStatus`**: Every production `BundleMetadata` construction now sets status explicitly, making the checkpoint intent visible at each call site
- **Design docs updated**: Bundle state machine, filter subsystem, BPA design, storage subsystem, and routing docs all reflect the new architecture

Closes #522


sylvain-pierrot (approved):
LGTM!

PR #528, opened 2026-05-01
Add zero-copy support to Editor and use in Parser

### LINKED ISSUE #147: Allow in-place modification of Bundles with minimal memory copies
Hi again :).

As far as I can tell, the `hardy_bpv7::bundle::Bundle` struct does not carry the actual payload data. Instead, one has to carry a serialized form of the bundle next to the `Bundle` struct. In order to modify the bundle's structural information or data, one has to go through `Builder` and serialize a complete new bundle.

If I am reading this correctly, this means that the (potentially large) bundle payload data has to be copied to a new memory location completely, even though it could be re-used if the previous version of the bundle is to be discarded.

This seems particularly relevant when implementing forwarding (RFC 9171, §5.4) in memory-constrained environments, as Step 4 of that algorithm has up to three modifications which need to happen to bundles. So every transmission would cause a complete copy of the bundle, potentially just before copying it again into the CLAs transmission buffers.

That seems rather inefficient. Am I missing something here? If I'm not missing anything here, do you have plans or ideas on how to modify the bpv7 crate to avoid this extra copy of data? Do you have a different judgement on how frequent those copy cases truly are?

Thanks for your time :).

### LINKED ISSUE #526: Refactor block serialization to use ToCbor trait
# Problem

Block serialization is currently split across multiple ad-hoc emit() methods:

- Block::emit() — manually constructs CBOR array (type, number, flags, CRC type, data, CRC value), calls append_crc_value, tracks extent/data ranges
- PrimaryBlock::emit() — similar manual CBOR construction with different fields, same CRC pattern
- Bundle::emit_primary_block() — wraps PrimaryBlock::emit(), emits into array, updates self.blocks[0] with extent

These methods duplicate the CRC composition pattern and bypass the standard ToCbor trait that the CBOR crate provides.

## Proposed change

Implement ToCbor for PrimaryBlock and Block (or a wrapper that carries block number), with CRC calculation as part of the trait impl via append_crc_value.

### Before:
```rust
// Builder
bundle.emit_primary_block(a)?;

// Block emission
block.emit(block_number, data, array)?;
```

### After:
```rust
// Builder
let extent = a.emit(&primary_block);

// Block emission
let extent = a.emit(&block_with_data);
```

## Benefits

- append_crc_value becomes an internal detail of ToCbor, not exposed at call sites
- Block serialization uses the standard trait — consistent with all other CBOR types
- Simplifies Builder, Editor, and parse code
- Extent tracking separated from serialization (caller reads extent from emit() return value)

## Call sites affected

- bpv7/src/builder.rs — bundle.emit_primary_block(a)
- bpv7/src/editor.rs — bundle_out.emit_primary_block(a) (BundleUpdate path), template.build() (extension blocks)
- bpv7/src/bundle/parse.rs — PrimaryBlock::emit(bundle) (canonicalization)
- bpv7/src/block.rs — Block::emit(), Block::copy_whole()

## Considerations

- Block number is not part of Block struct — ToCbor impl may need a wrapper like BlockWithNumber { block, number, data } or the number could be passed separately, or perhaps ToCbor<(u64,&Block)>
- copy_whole() (raw byte copy for unchanged blocks) remains separate — it bypasses serialization entirely
- CRC type is already a field on both Block and PrimaryBlock, so append_crc_value has everything it needs

ricktaylor:
Closes: #147 
Closes: #526 

# Editor: rebuild_bundle() returns (Bundle, Vec<Chunk>)

- rebuild_bundle(): Returns updated Bundle with correct block extents and BPSec coverage, plus Vec<Chunk> for zero-copy output
- No clone: Bundle constructed field-by-field during emission — convenience fields (previous_node, age, hop_count) parsed inline from block payloads. Keep blocks copy from original, Update blocks re-parse, deleted blocks stay None
- BPSec correctness by construction: Public API guards reject BIB/BCB on push_block/insert_block/update_block (Error::SecurityBlock); alloc_block/update_block_inner/remove_block_inner are pub(crate) for Signer/Encryptor
- Security metadata overrides: set_bib_target()/set_bcb_target() let Signer/Encryptor set coverage explicitly — no reparse fallback needed

# Chunk-based rebuild (zero-copy)

- Chunk type: Unchanged(Range<usize>) references original source data, New(Box<[u8]>) holds freshly encoded bytes
- rebuild(): Returns Vec<Chunk>; callers flatten explicitly
- Chunk::flatten(): Concatenates chunks with 0x9F/0xFF CBOR array wrapper
- Chunk::flatten_inplace(): Modifies source buffer in place — forward or backward pass depending on whether output grows; single scan pass + single write pass
- Chunk::assemble(): Orders chunks optimally — primary first, payload last, Unchanged extensions sorted by source position, New chunks slotted into gaps of matching size, adjacent Unchanged ranges merged
- Block extent tracking: assemble() sets block extents relative to flattened output (offset 1 for 0x9F)

# Primary block handling

- set_canonical_primary(): pub(crate) method for parser to provide pre-encoded canonical primary block bytes
- PrimaryBlock::emit() called directly instead of Bundle::emit_primary_block() — the latter removed
- build_to_vec() on builder::BlockTemplate — encodes extension blocks to standalone bytes without needing a CBOR Array encoder
- Block::copy_whole() and old array-based build_block() removed — build_chunk() is the sole path

# Signer/Encryptor

- rebuild_editor() extracted as shared internal method
- Both expose rebuild() (returns Vec<Chunk>) and rebuild_bundle() (returns (Bundle, Box<[u8]>))
- Set bib/bcb overrides explicitly during signing/encryption

# parse.rs finish() migration

- Rewritten to use Editor instead of manual CBOR emission — emit_block() removed
- Primary block: set_canonical_primary() with pre-emitted bytes
- Block removals: remove_block_inner() with panic on failure
- Non-canonical blocks: update_block_inner().with_data().rebuild()
- Returns Vec<Chunk> propagated through RewrittenBundle::Rewritten and CheckedBundle

# BPA: Zero-copy bundle rewriting at call sites

- dispatch.rs, restart.rs, local.rs: Bytes::try_into_mut() + flatten_inplace() when buffer exclusively owned; fallback to flatten() when shared
- forward.rs: rebuild() + try_into_mut / flatten_inplace for extension block updates
- adu_reassembly.rs: Same pattern for fragment reassembly
- echo-service, pipeline.rs: Same pattern for source/destination swap
- filter/chain.rs: flatten_inplace on filter-provided Vec<u8> for canonicalization

# Tests

- rebuild_bundle() roundtrip tests verify Bundle matches fresh parse (extents, block fields, BPSec coverage)
- flatten_inplace tests verify byte-identical output to flatten across no-op, destination change, block add, block remove
- All 7-implementation interop tests passing
- Significant performance improvement observed

### DIFF
--- a/bpa/src/dispatcher/admin.rs
+++ b/bpa/src/dispatcher/admin.rs
@@ -3,7 +3,7 @@ use hardy_bpv7::status_report::AdministrativeRecord;
 
 impl Dispatcher {
     #[cfg_attr(feature = "instrument", instrument(skip_all,fields(bundle.id = %bundle.bundle.id)))]
-    pub(super) async fn administrative_bundle(&self, mut bundle: bundle::Bundle, data: Bytes) {
+    pub(super) async fn administrative_bundle(&self, mut bundle: bundle::Bundle) {
         metrics::counter!("bpa.admin_record.received").increment(1);
 
         // This is a bundle for an Admin Endpoint
@@ -16,6 +16,14 @@ impl Dispatcher {
                 .drop_bundle(bundle, ReasonCode::BlockUnintelligible)
                 .await;
         }
+
+        let Some(data) = self.load_data(&bundle).await else {
+            if !bundle.has_expired() {
+                // Bundle data was deleted while queued - not reaped
+                self.drop_bundle(bundle, ReasonCode::DepletedStorage).await;
+            }
+            return;
+        };
 
         let payload_result = {
             let key_source = self.key_source(&bundle.bundle, &data);
--- a/bpa/src/dispatcher/admin.rs
+++ b/bpa/src/dispatcher/admin.rs
@@ -3,7 +3,7 @@ use hardy_bpv7::status_report::AdministrativeRecord;
 
 impl Dispatcher {
     #[cfg_attr(feature = "instrument", instrument(skip_all,fields(bundle.id = %bundle.bundle.id)))]
-    pub(super) async fn administrative_bundle(&self, mut bundle: bundle::Bundle) {
+    pub(super) async fn administrative_bundle(&self, bundle: bundle::Bundle) {
         metrics::counter!("bpa.admin_record.received").increment(1);
 
         // This is a bundle for an Admin Endpoint
@@ -17,11 +17,7 @@ impl Dispatcher {
                 .await;
         }
 
-        let Some(data) = self.load_data(&bundle).await else {
-            if !bundle.has_expired() {
-                // Bundle data was deleted while queued - not reaped
-                self.drop_bundle(bundle, ReasonCode::DepletedStorage).await;
-            }
+        let Some((mut bundle, data)) = self.load_data_or_drop(bundle).await else {
             return;
         };
 
--- a/bpa/src/dispatcher/dispatch.rs
+++ b/bpa/src/dispatcher/dispatch.rs
@@ -3,6 +3,255 @@ use futures::{FutureExt, join, select_biased};
 use hardy_bpv7::status_report::ReasonCode;
 
 impl Dispatcher {
+    /// Entry point for bundles received from CLAs.
+    ///
+    /// Parses the CBOR-encoded bundle, validates the format, stores bundle data,
+    /// inserts initial metadata with `New` status, and queues for ingestion.
+    ///
+    /// # Bundle State
+    ///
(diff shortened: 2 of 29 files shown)


sylvain-pierrot (approved):
Nice improvements! LGTM

PR #533, opened 2026-05-07
Replace flume in hardy-bpa with hardy-async channel + StreamIn trait

ricktaylor:
## Summary

- Removes `flume` from `hardy-bpa`'s production dependencies, unblocking one of the two listed `no_std` blockers in `bpa/src/lib.rs`.
- Introduces `hardy_async::channel` (drop-to-close MPMC) and `hardy_async::closeable` (explicit close via shared `CancellationToken`) as runtime-agnostic channel wrappers — currently flume-backed under `cfg(feature = "std")`, ready for alternative backends in future no_std implementations.
- Refactors the `MetadataStorage` and `BundleStorage` trait surfaces to take `&dyn StreamIn<T>` instead of `flume::Sender<T>`, decoupling every storage backend from any specific channel implementation.

## Motivation

`hardy-bpa`'s `lib.rs` previously listed two blockers preventing `no_std` compilation: `flume` and `metrics`. This PR removes the `flume` blocker. The `metrics` blocker is captured as a separate issue (planned `hardy-metrics` facade crate).

Beyond unblocking `no_std`, the refactor cleans up two long-standing awkwardnesses:

1. The hybrid storage channel (`bpa/src/storage/channel.rs`) used `flume::bounded::<Option<Bundle>>(cap)` with a `None` sentinel as the close signal. The `Option<Bundle>` leaked into the public storage `Receiver` type alias and into every consumer's match arms.
2. The storage trait methods (`poll_pending`, `recover`, etc.) took a concrete `flume::Sender<T>` parameter, which forced every backend crate (sqlite, postgres, localdisk, s3, in-mem) to depend on flume and follow flume's API conventions even when their internal model (e.g. localdisk's `spawn_blocking` walk) didn't fit cleanly.

### Two new wrappers in `hardy-async`

| Module | Close semantics | Used by |
|---|---|---|
| `hardy_async::channel` | Drop-to-close (all senders gone → receiver disconnects) | Most BPA call sites |
| `hardy_async::closeable` | Explicit `Sender::close()` via shared `CancellationToken`; receiver drains buffer then disconnects | BPA's hybrid storage channel |

Both expose `bounded`/`unbounded` constructors, `try_send`/`send`/`recv`, `len`/`is_empty`, full `Debug`/`Display`/`core::error::Error` impls on the error types, and comprehensive tests covering close semantics, drain ordering, multi-producer behaviour, and disconnection paths.

### `StreamIn<T>` trait surface for storage

A new `pub trait StreamIn<T>: Send + Sync` with a single async `send` method replaces the `Sender<T>` alias on storage trait methods. Three adapter shapes:

- `ChannelStreamIn<T>` (in `bpa::storage`) wraps a `hardy_async::channel::Sender` for the BPA's own call sites.
- A small `flume::Sender` adapter inside `localdisk-storage`'s `recover()` lets the existing `spawn_blocking` walk code stay unchanged behind a bridge pump task.
- A `VecSink<T>` shared test helper in `tests/storage/src/lib.rs` collects items into a `Mutex<Vec<T>>` for assertions.

See `bpa/docs/storage_subsystem_design.md` "Streaming results via `StreamIn<T>`" for the design rationale, naming choice (vs `Sink`, vs `futures::Stream`), and connection to the planned streaming-chunks direction.

### Other cleanup

- `Peer::close` made sync (body had no awaits left after migration).
- `closeable::recv` switched from `tokio::select!` to `futures::select_biased!` — runtime-agnostic, partially decouples closeable from the `tokio` feature (full decoupling needs a runtime-agnostic `CancellationToken`, tracked separately).
- `bpa::std` feature now correctly forwards to `hardy-async/std` (was silently broken before — a `--no-default-features --features std` build would have failed).
- `bpa-server::instrument` feature now forwards to `hardy-async/instrument`.
- New `bpa::critical-section` feature forwards to `hardy-bpv7/critical-section` (the previous instruction in BPA's README to "enable `critical-section` on `hardy-bpa`" was broken — no such feature existed).
- Several `std::*` references in BPA replaced with `core::*` / `alloc::*` / `hashbrown::*` / `hardy_async::sync::spin::*` equivalents, bringing BPA closer to actually compiling under `no_std` once the `metrics` blocker is also resolved.
- `#[inline]` annotations added across hot-path methods in `hardy-async` (channel/closeable send, recv, From impls, forwarding  methods on `BoundedTaskPool` and `Notify`).

## Affected crates

- `hardy-async` — new `channel` and `closeable` modules; `Once::wait`, `futures` dep, `#[inline]` annotations
- `hardy-bpa` — Pattern A (closeable hybrid channel), Pattern B (StreamIn trait surface), no_std cleanup, feature flag fixes
- `hardy-bpa-server` — `instrument` feature forwarding
- `hardy-sqlite-storage` — trait impls migrated to `StreamIn`, flume dropped from dev-deps
- `hardy-postgres-storage` — trait impls migrated to `StreamIn`
- `hardy-localdisk-storage` — `recover()` rewired with internal flume - bridge pump pattern, walk code unchanged
- `hardy-s3-storage` — `recover()` migrated to `StreamIn`
- `tests/storage` — shared `VecSink<T>` test helper, all suite call sites migrated, flume dropped from deps


bpa/src/storage/channel.rs
```diff
@@ -21,17 +47,27 @@
...
+    bundle::{Bundle, BundleStatus},
+    stream::ChannelSender,
+};
+
+use super::store::Store;
+
+/// Receiver handle for bundles drained from a hybrid storage channel.
+/// `recv()` returns `Err(RecvError::Disconnected)` after the buffer
+/// drains once the channel has been closed.
+pub(crate) type Receiver = hardy_async::closeable::Receiver<Bundle>;
```
sylvain-pierrot: What is the purpose of this alias? I’d rather remove the alias as this hides the concrete type.

bpa/src/storage/channel.rs
```diff
@@ -80,7 +116,7 @@ impl ChannelState {
 /// Shared state between Sender and the background poller task.
 struct Shared {
     state: AtomicUsize,
-    tx: flume::Sender<Option<Bundle>>,
+    tx: hardy_async::closeable::Sender<Bundle>,
```
sylvain-pierrot, line 113: Here we don’t have alias, and therefore more understandable 🙂

bpa/src/stream.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[async_trait]
+impl<T: Send + 'static> Sender<T> for ChannelSender<T> {
+    async fn send(&self, item: T) -> core::result::Result<(), SendError<T>> {
+        self.0
+            .send(item)
+            .await
+            .map_err(|hardy_async::channel::SendError(item)| SendError(item))
+    }
+}
```
sylvain-pierrot, line 59: I’m not entirely convinced here. `hardy_async::channel::Sender` is already channel agnostic, why do we need to add this new abstraction? We’re starting to have multiple abstraction layers for channels. I’m worried about the fact that we have now three Sender types, and the need to track the origin of a `Sender`. Additionally, the `trait Sender` has just one implementation, do you plan to add more?

ricktaylor: Concrete numbers after the collapse — there are now two Sender types in this area, and the trait has two kinds of implementor:
  
Sender types
 - stream::Sender<T> — the trait (the storage-backend SPI).
 - hardy_async::channel::Sender<T> — the concrete channel sender, which now implements the trait directly. The ChannelSender adapter is gone, so we're down from three to two.

Implementations of stream::Sender<T>
 1. hardy_async::channel::Sender<T> — the production sink (bpa/src/stream.rs).
 2. VecSink<T> — the test collector: in the shared conformance harness (tests/storage/src/lib.rs) and in the #[cfg(test)] modules of localdisk-storage and sqlite-storage.

So it isn't a single-impl abstraction. VecSink (a Mutex<Vec<T>>) is what lets each backend's poll_*/recover be driven and asserted on with no channel and no receiver task — which is exactly what you can't express as a channel::Sender. The trait stays "one sink per consumer," and the channel sender is no longer wrapped.

sylvain-pierrot: Than you for you reply and the changes. We are in the right direction 🙂 

I have another concern though. I like the streaming abstraction in `stream.rs` however keeping the naming `Sender` is not really adapted since it refers to a channel. If my understanding is correct, we would like to have a stream abstraction without caring if under the hood the backend is a channel or whatever. Hardy Async is definitely the right implementation, however for stream abstraction in the BPA, we shouldn’t use the channel terminology, and use the Read/Write streaming terminology design.

ricktaylor: Not necessarilly - I tried a number of different names, and Sender/Receiver (both async traits) seemed the best name - everything else was more confusing

sylvain-pierrot: What do you think about the std crate streaming naming? Read/Write

Sender/Receiver gives a strong feeling of manipulating a resource, and streaming is “reading a chunk somewhere” and “writing a chunk somewhere”

ricktaylor: Reader/Writer implies something is being "Written" and it might not be - it's just being "passed to the next stage" - so no underlying implication of persistence

sylvain-pierrot: The streaming abstraction lives at the boundaries of the BPA between a CLA/BPA, BPA/Storage, etc. At the boundary we read and write. Then the item (a chunk) flows in the BPA, and the chunk itself is not a stream anymore.

ricktaylor: Aah.. but it's a sender or receiver of chunks - just piped through

sylvain-pierrot: The idea with the trait is to not know about the sender/receiver, otherwise we should use the channel directly :/

I don’t have the picture yet for how a chunk flows through the BPA, but if we’re using channels for that, this is another concern

ricktaylor: We're going to use 'channel-like' things

sylvain-pierrot: Chunk flows and streaming abstraction are two different things (they are not related), so even if we’re going to do channels “explicitly" for chunk flows, that should not impact the streaming abstraction by calling this sender/receiver, shouldn’t it?

sylvain-pierrot (dismissed):
Great improvements!

bpa/src/stream.rs
ricktaylor: Aaah... there are 2 changes that will come with "streaming":  Processing blocks as 'chunks', and durable queues - both need the same abstractions, hence the more generic names

sylvain-pierrot (approved):
LGTM

PR #535, opened 2026-05-11
Fix/bpsec interop

sylvain-pierrot:
# BPSec Bug Fixes

Bugs found and fixed during BPSec interoperability testing.

## Decrypted BCB targets not rewritten in Full mode

**Tests:** 2.6, 2.8

**Symptom:** After `RewrittenBundle` decrypted a BCB target (e.g., an encrypted BIB), the rebuilt output still contained the ciphertext. Subsequent operations that tried to parse the block's data (like `remove_integrity`) failed with `IncorrectType("Array", "Byte String")`.

**Root cause:** Three issues in `bpv7/src/bundle/parse.rs`:

1. `decrypt_bcbs` stored plaintext in `decrypted_data` (transient), but never told `finish()` to write it to the output. The `noncanonical_blocks` map (used by `finish()` for rewriting) was not updated.

2. `reduce_bcbs` only cleaned up BCB operations for blocks in `blocks_to_remove` (deleted blocks). It did not handle decrypted targets: the block's `bcb` field was never cleared, and the BCB's operation for the decrypted target was never removed. Empty BCBs were left in the bundle.

3. `decrypted_data.clear()` ran before `reduce_bcbs`, so by the time cleanup ran, the plaintext was already gone.

**Fix:**

- Added `decrypted_blocks: SmallVec<[(u64, u64); 8]>` to `BlockParse` to track successfully decrypted targets alongside their BCB block numbers.
- `decrypt_bcbs` records each successful decryption in `decrypted_blocks`.
- `reduce_bcbs` (Full mode only) now handles decrypted targets: clears `bcb` on the target block, stores the plaintext from `decrypted_data` into `noncanonical_blocks`, removes the target from the BCB's operation set, and removes empty BCBs.
- Moved `decrypted_data.clear()` to after `reduce_bcbs` so plaintext is available during cleanup.

## Non-payload BCB decrypt failure discarded entire bundle

**Test:** 37.1

**Symptom:** When a BCB targeting a non-payload block (e.g., bundle-age) could not be decrypted (wrong key), Hardy discarded the entire bundle.

**Root cause:** `decrypt_bcbs` returned `Err` on any `DecryptionFailed`, regardless of whether the target was the payload.

**RFC 9172 Section 5.1.1** distinguishes:
- Payload: "the bundle MUST be discarded"
- Non-payload: "the associated security target and all security blocks associated with that target MUST be discarded" (bundle survives)

**Fix:** In `decrypt_bcbs`, catch `DecryptionFailed` for non-payload targets and route them to `blocks_to_remove` (like the `NoKey` path), while still failing fatally for payload targets.

## Fragment validation rejected offset=0, total=0

**Test:** 56.1

**Symptom:** Hardy rejected the PICS fragment bundle with `InvalidFragmentInfo(0, 0)`.

**Root cause:** `primary_block.rs` enforced `offset < total_adu_length`. With both zero, `0 >= 0` triggered the error.

**Fix:** Changed to `offset > total_adu_length`. RFC 9171 does not explicitly forbid a zero-length ADU fragment.


bpv7/src/editor.rs
```diff
@@ -1011,10 +1011,9 @@ impl<'a> Editor<'a> {
                 // We must decrypt such BIBs and remove the signature.
 
                 // Handle BIBs within this same BCB's targets.
-                // Note: BCB-AES-GCM (RFC 9173) cannot have multiple targets due to IV
-                // uniqueness requirements, so this code path is for future security
-                // contexts (e.g., COSE-based) that may support multi-target BCBs.
-                if opset.can_share() {
```
ricktaylor, line 1017: No - don't remove this check in the Editor...  we MUST NOT produce invalid shared targets.   We can however parse them, but we cannot build them!

ricktaylor (changes requested):
One quick comment

bpv7/src/editor.rs
sylvain-pierrot: this is the behavior that I wanted to targeted

sylvain-pierrot: For the build we enforce here: bpv7/src/bpsec/encryptor.rs, lines 201-242.

sylvain-pierrot: Hold on, your point is only for parsing or also editing a multi targets?

ricktaylor: Only editing - if someone else creates bundles in violation of 9173, we should at least be able to process them - we should probably have an explicit filter one day that checks this at a higher level.

bpv7/tools/Cargo.toml
```diff
@@ -5,12 +5,15 @@ edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
-description = "CLI tools for BPv7 bundle operations"
+description = "CLI tools and library for BPv7 bundle operations"
 publish = false
 
+[lib]
```
ricktaylor: I wouldn't bother making this a library at this time - I think we need to decide what we do with the bpv7 crate in general first.  Perhaps compare.rs actually moves into the crate itself?

sylvain-pierrot: I thought about this possibly as well. Let’s move `compare` into `bpv7` then?

bpv7/src/editor.rs
sylvain-pierrot: So for processing/editing them, I think this is the right change, isn't it?

tests/bpsec/src/lib.rs
ricktaylor, line 1: I would merge this with th pics_defaults.rs - it's very specific to that test suite

ricktaylor (changes requested):

tests/bpsec/src/lib.rs
sylvain-pierrot: Moved!

ricktaylor (approved):
LGTM

PR #541, opened 2026-05-12
Feat/bundle compare

sylvain-pierrot:
## Description

Bundles in BPv7 don't have a single canonical byte representation. The RFCs allow CBOR encoding choices, arbitrary block numbering, block reordering, and security block internal array ordering that all produce different bytes for the same semantic content. This makes byte-level comparison unreliable for conformance testing and interoperability validation.

This PR adds a compare module to the `bpv7` crate that compares bundles at the parsed level, handling these encoding freedoms transparently. 

See [`bpv7/docs/bundle_compare.md`](bpv7/docs/bundle_compare.md) for the full design and rationale.

## Usage

```rust
use hardy_bpv7::cmp::compare_bundles;

let diffs = compare_bundles(&data_a, &data_b)?;
```


bpv7/tests/cmp.rs
ricktaylor, line 1: I think this might be best near cmp.rs in the main bpv7 crate rather than in the tools subdirectory

ricktaylor (changes requested):
Really good - except I think the tests should be in the main bpv7 crate as well as cmp.rs

bpv7/tests/cmp.rs
sylvain-pierrot: I’m doing the change

bpv7/src/editor.rs
```diff
@@ -1011,10 +1011,9 @@ impl<'a> Editor<'a> {
                 // We must decrypt such BIBs and remove the signature.
 
                 // Handle BIBs within this same BCB's targets.
-                // Note: BCB-AES-GCM (RFC 9173) cannot have multiple targets due to IV
-                // uniqueness requirements, so this code path is for future security
-                // contexts (e.g., COSE-based) that may support multi-target BCBs.
-                if opset.can_share() {
```
ricktaylor, line 1017: No - please don't remove this check!

ricktaylor (changes requested):
Just put the shared IV check back in and we're good!

bpv7/src/editor.rs
sylvain-pierrot: Undo ;)

ricktaylor (approved):
LGTM!

PR #542, opened 2026-05-12
Add TaggedMarker for low-level CBOR dispatch and skip_value for allocation-free skipping

ricktaylor:
Summary

- Introduce Marker and TaggedMarker — a new low-level CBOR decode API that exposes the item head (major type, length/value, tags) without requiring a closure or materialising indefinite-length chunk lists. Implements FromCbor so it composes with
parse::<TaggedMarker>(data).
- Add top-level skip_value(data, max_recursion) — walks past any CBOR item (including nested arrays/maps) without constructing Value, Series, or chunk Vecs. Replaces the parse_value + Value::skip pattern throughout, deprecating the old Value::skip method.
- Refactor internals to use TaggedMarker — parse_value, parse_array, parse_map, Series::parse_array, Series::parse_map, and the FromCbor impls for primitives (u64, i64, f64, bool) now decode via TaggedMarker instead of duplicating the major-type dispatch. This
removes ~100 lines of redundant match arms and fixes a parse_data_minor return-value bug (was returning the wrong end offset for definite-length strings).
- Minor fixes: Series::at_end uses is_multiple_of() for partial-map detection; shortest &= s replaces shortest = shortest && s throughout.
- bpv7 callers updated: parse.rs test and eid/mod.rs Display impl switched to the new skip_value / skip_to_end APIs.

Test plan

- Existing cbor decode test suite passes (including malformed/recursion tests, updated to use skip_value)
- bpv7 test suite passes (unknown-block insertion test updated)
- Fuzz corpus regression (cbor + bpv7)


ricktaylor:
# PR Update — Review Feedback Addressed

## Bug fixes

HIGH — skip_value skipped half the items in a definite-length map (decode.rs)
The joint Marker::Array(Some(count)) | Marker::Map(Some(count)) arm looped count times, but Marker::Map(Some(n)) carries pair count, not item count. Maps with N pairs were skipping only N items, leaving N items unread and corrupting the offset for everything that
followed.

LOW — Series::skip_to_end infinite-looped for D=0 (Sequence) (decode_seq.rs)
The inner for _ in 0..D ran zero times for Sequence, so offset never advanced. Latent today (no caller invokes it on a Sequence), but a regression from the previous parse_value-based implementation. Changed to "always skip one, then conditionally more" so D=0/1 skip
one item per outer iteration and D=2 skips a key/value pair (preserving PartialMap detection via at_end).

LOW — skip_value accepted malformed indefinite maps with odd item counts (decode.rs)
The indefinite Map/Array arm walked items until 0xFF without parity-checking. An indefinite map like BF 01 02 03 FF (3 items, 1.5 pairs) was silently accepted; the old parse_value path would have raised Error::PartialMap.

Three regression tests added covering each bug plus their adjacent edge cases (sentinel after a definite map; nested map in array; empty/odd/even indefinite maps; empty sequence).

## Refactoring

Consolidated count handling into Series::try_new (decode_seq.rs)
The 8 call sites that built Array/Map from a CBOR head all duplicated the try_from(usize) + (for maps) checked_mul(2) boilerplate — and forgetting the * 2 was the root cause of the HIGH bug. New pub(super) Series::try_new(data, count: Option<u64>, offset) does the
conversion in one place using D.max(1) so Map (D=2) doubles, Array (D=1) is identity, and Sequence (D=0) — which never carries a count — falls through harmlessly. The * 2 for maps now lives at exactly one source line crate-wide.

Collapsed skip_value Array/Map dispatch via Series::skip_to_end (decode.rs)
Four arms (definite/indefinite × Array/Map) reduced to two. Series::at_end already discriminates definite vs indefinite and handles PartialMap detection, so the duplicated logic disappeared. Recursion remains well-formed (decode::skip_value → Series::skip_to_end →
Series::skip_value → decode::skip_value); max_recursion is decremented once per container level as before.

## Polish (other items from review)

- Marker doc (decode.rs): corrected misleading "head byte, length prefix, and payload bytes are all consumed" claim for definite-length strings — the payload remains in the buffer; only head + length prefix are consumed. Also fixed Some(range) → Some(len) references
that predated the variant becoming Option<u64>.
- TaggedMarker::Display (decode.rs): removed .to_string() on a &str literal — now uses the borrow directly with write!.
- Marker (decode.rs): added #[derive(Debug, Clone, PartialEq)]. Doc notes the IEEE-754 NaN gotcha for Float.
- Marker::End offset == 1 guard (decode.rs): added a comment explaining the guard discriminates untagged break codes (valid End) from tagged ones (rejected as InvalidSimpleType(31)).
- Value::type_name (decode.rs): rewrote to use format! with a &str prefix instead of String + &str chains — one allocation per call instead of one-or-two. Output strings unchanged.

## Performance

- Hot-path parse_array / parse_map (used heavily in BPv7 bundle parsing) and their Series:: counterparts kept as direct TaggedMarker dispatch — try_new is invoked but no closure-through-parse_value indirection.
- All other perf wins from the original PR (primitive FromCbor impls bypassing parse_value, zero-alloc skip_value) are preserved.

## Behaviour change worth noting

parse_array / parse_map on a bare 0xFF byte previously returned Error::IncorrectType("Array", "End of sequence marker"); this is unchanged in this revision (we kept direct dispatch). If we later collapse them to delegate through parse_value, the error becomes
Error::InvalidSimpleType(31) — arguably more correct (a bare break is malformed CBOR, not a type mismatch), but worth flagging if any caller pattern-matches on IncorrectType.

cbor/tests/decode.rs
sylvain-pierrot, line 1: I think you can move this file in `/tests`?

ricktaylor: Done!

sylvain-pierrot: I meant in cbor/tests

sylvain-pierrot (dismissed):
Amazing! Looks really clean. Approved

sylvain-pierrot (approved):

PR #543, opened 2026-05-13
Feat/key management

### LINKED ISSUE #297: Add config support for key material to the BPA
As with the `bundle` subcommands, we need the ability to supply keys to the BPA.  Options include:

* A JWKS set as part of the configuration.
* A path to file or directory directory with JWK and JWKS in, plus a watcher.
* Some kind of API access, so a `KeyStore` trait impl can be passed to the BPA on start up.

The perfect solution would be a combination of all 3:
1. Implement a way to associate a `KeyStore` trait impl with the dispatcher
2. Implement a `KeyStore` trait impl for config
3. Implement a `KeyStore` trait impl for a file watcher
4. Implement a public API on BPA to layer multiple impls.

sylvain-pierrot:
## Description

> Close #297.

The BPA has no way to provide cryptographic keys for BPSec operations. Bundles arriving with BIB or BCB blocks are dropped with `NoKey`.

This PR adds key management to the BPA, enabling verification and decryption of inbound BPSec-protected bundles (RFC 9172/9173). Keys are loaded from a JWK Set file (RFC 7517) and bound to peers via EID patterns (draft-ietf-dtn-eid-pattern). Each binding specifies a security role (verifier, acceptor, or source) per RFC 9172 Section 2.5. The key file can optionally be watched for changes and reloaded at runtime.

Keys and bindings can also be added or removed at runtime via the `KeyStore` API.

## Providing keys

**1. Create a JWK Set file** (e.g. `/etc/hardy/keys.jwks`) with your pre-shared keys:

```json
{
  "keys": [
    {
      "kid": "fleet-hmac",
      "kty": "oct",
      "k": "hJtXIZ2uSN5kbQfbtTNWbpdmhkV8FJG-Onbc6mxCcYg",
      "alg": "HS256",
      "key_ops": ["sign", "verify"]
    },
    {
      "kid": "node42-aes",
      "kty": "oct",
      "k": "hJtXIZ2uSN5kbQfbtTNWbpdmhkV8FJG-Onbc6mxCcYg",
      "alg": "A256KW",
      "enc": "A256GCM",
      "key_ops": ["encrypt", "decrypt", "wrapKey", "unwrapKey"]
    }
  ]
}
```

Save it with restricted permissions (`chmod 0600`). Each key needs a `kid`, `kty: oct` (symmetric), and `key_ops`.

**2. Add a `bpsec` section** to the BPA server config:

```yaml
bpsec:
  keys-file: /etc/hardy/keys.jwks
  watch: poll
  bindings:
    - match: "ipn:*.*"
      role: acceptor
      keys: ["fleet-hmac", "node42-aes"]
    - match: "ipn:0.2.*"
      role: verifier
      keys: ["fleet-hmac"]
    - match: "ipn:0.3.*"
      role: source
      keys: ["fleet-hmac", "node42-aes"]
```

Each binding maps an EID pattern to a role and a set of keys (referenced by `kid`). The BPSec backend selects the right key by matching the required operation against each key's `key_ops` field.

When multiple bindings match, the most specific pattern wins (`ipn:0.2.*` overrides `ipn:*.*`).

## Security roles

Each binding specifies the BPA's role for security blocks from/to that peer (RFC 9172 Section 2.5):

| Role | Direction | BIB behavior | BCB behavior |
|------|-----------|--------------|--------------|
| `verifier` | Inbound | Verify, keep | Pass through |
| `acceptor` | Inbound | Verify + remove | Decrypt + remove |
| `source` | Outbound | Add BIB at egress | Add BCB at egress |

When the BPA is the bundle destination, it acts as acceptor for remaining BCBs regardless of the binding's role (RFC 9172 Section 5.1.1).

Default role when omitted: `verifier`.

> Note: role-based processing (acceptor block removal, source egress filter) is not yet implemented. Currently all roles verify BIBs and decrypt BCBs in memory without removing security blocks.

## File watching

The `watch` field enables automatic key reload when the key file changes:

| Value | Behavior |
|-------|----------|
| `native` | OS-native file events (inotify/kqueue). Fast, but does not work across Docker bind mounts. |
| `poll` | Periodic filesystem polling (~2s). Works everywhere including Docker. |
| *(omitted)* | Watching disabled. Keys are loaded once at startup. |

If a reload fails (malformed JSON, invalid keys), the previous keys are kept and an error is logged.

The same watcher is also used by static routes (`static-routes.watch`), which now accepts the same `native`/`poll` values instead of a boolean.


ricktaylor (commented):
Okay - immediate comment:  Can we move the "fix watcher for docker and make re-useable" into a seperate PR, possibly moving the functionality into hardy-async as it's a common and reuseable pattern?  It would keep this PR really focussed

ricktaylor:
For the key selection in the config - give we have "key_ops" on the actual keys, and the BPSec backend will select and appropriate key based on key_ops, why do we need to specify "integrity_key" and "confidentiality key"?  Surely we can say "This is the set of keys"

ricktaylor (changes requested):
Hmm...  Although I like the KeyPatternSource and the config etc - I really don't want to lose the Keys::Registry - It's important long-term to be able to register multiple key providers.

Can we make the new code implement a KeyProvider and add to the registry, rather than replace the registry with the new code.

sylvain-pierrot:
> Okay - immediate comment: Can we move the "fix watcher for docker and make re-useable" into a seperate PR, possibly moving the functionality into hardy-async as it's a common and reuseable pattern? It would keep this PR really focussed

I agree with making this reusable. However, what’s the purpose of the crate hardy-async? This is not really async.
And I’ll make the change before this PR. I don’t want to have this code waiting for the `waiter` after merging.

sylvain-pierrot:
> For the key selection in the config - give we have "key_ops" on the actual keys, and the BPSec backend will select and appropriate key based on key_ops, why do we need to specify "integrity_key" and "confidentiality key"? Surely we can say "This is the set of keys"

I probably got lost while thinking about a fast key lookup. I like your comment though. And so, here is the config:

```yaml
  bindings:
    - match: "ipn:*.*"
      role: acceptor
      keys: ["fleet-hmac”, "node42-aes”]
```

Let me know if that works for you. 

With a bit more of context, the key selection looks like: pattern → binding → filter keys by key_ops + key_algorithm → one key.

sylvain-pierrot:
> Hmm... Although I like the KeyPatternSource and the config etc - I really don't want to lose the Keys::Registry - It's important long-term to be able to register multiple key providers.
> 
> Can we make the new code implement a KeyProvider and add to the registry, rather than replace the registry with the new code.

I see here multiple topics:

1. We lose runtime registration, and I agree with making this back. 
2. I don’t see the utility of having multiple implementations of `trait KeyProvider`, for me this add just a runtime cost. And also we already have a trait from `bpv7` -> `KeySource`.
3. We could just keep `PatternKeySource`, and deal with `keys` and `bindings` (structure fields) for adding and removing keys.
4. Naming: I don’t like the name “Registry” when it’s a half registry behavior.
5. If a genuinely different provider appears later, adding the registry behavior is a small change: wrap the `ArcSwap<PatternKeySource>` into `ArcSwap<Vec<Arc<dyn KeySource>>>`.

bpa-server/src/config/static_routes.rs
ricktaylor, line 1: Is the watcher change part of the watcher PR?

bpa-server/examples/keys.jwks
ricktaylor, line 1: I think this should be an example file, rather than an actual config if we are committing it to the repo

ricktaylor (changes requested):

bpa-server/src/config/static_routes.rs
sylvain-pierrot: There is a small refactor for static-routes to use the watcher. And then, the watcher becomes generic as part of the other PR.

sylvain-pierrot: Sorry - I'll do smaller and specific PRs

bpa-server/examples/keys.jwks
sylvain-pierrot: Agreed! I'll make the change

sylvain-pierrot: Done!

ricktaylor:
I have cherry picked the critical BPSec validation onto main (already merged).  The update to KeyProvider is not needed now.  Not going to merge this

Issue #534, opened 2026-05-07
Introduce hardy-metrics crate to provide no_std-compatible metrics shims


ricktaylor:
## Context

`hardy-bpa`'s `lib.rs` lists `metrics` as the remaining `no_std` blocker (the `flume` blocker was removed by the `StreamIn` migration).

The upstream `metrics` crate is std-pinned and isn't on a path toward no_std support. BPA only uses a small subset of its API surface — four macros and one type — so we don't need to wait for upstream. A small facade crate can provide either the real `metrics` calls or no-op equivalents based on a feature flag.

(Compare with `tracing`, which is already no_std-compatible and needs no wrapping — `metrics` is the outlier in BPA's observability stack.)

## Proposal

Introduce a `hardy-metrics` crate with a `metrics` feature flag:

- **`metrics` on**: re-exports the upstream `metrics` crate verbatim.  Behaviour and runtime cost are identical to today.
- **`metrics` off**: provides no-op macro and type shims with the same shape. Calls compile to nothing.

The crate itself is `no_std`-clean. `hardy-bpa` depends on `hardy-metrics` instead of `metrics` directly, and BPA's existing `std` feature cascades down to `hardy-metrics/metrics`.

## API surface to provide

Concrete subset currently used in `hardy-bpa`:

| Macro / type | Used as |
|---|---|
| `counter!` | `counter!("bpa.foo").increment(1)` |
| `gauge!` | `gauge!("bpa.foo", "k" => v).increment(1.0)` / `.decrement(1.0)` / `.set(...)` |
| `describe_counter!` | `describe_counter!("bpa.foo", Unit::Count, "description")` |
| `describe_gauge!` | `describe_gauge!("bpa.foo", Unit::Bytes, "description")` |
| `Unit` (type) | `Unit::Count`, `Unit::Bytes` (variants used by BPA) |

`histogram!` and `describe_histogram!` are not currently used in BPA; add stubs for completeness but they're not required for this migration.

## Implementation sketch

The no-op side is small. Each macro expands to a unit struct that exposes the same chained methods as the real `metrics` types:

```rust
#![cfg_attr(not(feature = "metrics"), no_std)]

#[cfg(feature = "metrics")]
pub use metrics::{counter, gauge, histogram,
                    describe_counter, describe_gauge, describe_histogram,
                    Unit};

#[cfg(not(feature = "metrics"))]
mod noop {
    pub struct Counter;
    impl Counter {
[...]

PR #544, opened 2026-05-14
refactor(bpa): restructure bundle processing pipeline with standalone…

sylvain-pierrot:
… ingress


ricktaylor:
Yeah - this makes a lot of sense - but can we hold off on the "refactor and tidy up" pass until more of the streaming lands - I think we can fix the Cla trait under that work.

The metrics is going to need a bit more thought to support #534 and we could do the tidy up as part of the fix for that?

sylvain-pierrot:
> Yeah - this makes a lot of sense - but can we hold off on the "refactor and tidy up" pass until more of the streaming lands - I think we can fix the Cla trait under that work.
> 
> The metrics is going to need a bit more thought to support #534 and we could do the tidy up as part of the fix for that?

My thoughts for this Draft PR is to find a compromise between full-buffered bundle and  streaming. Introducing `ingress.rs` and `egress.rs` (or whatever naming), with applying the streaming workflow without streaming (ingress sequence: filter -> mutate -> route -> store) might be the progressive transition without getting lost.

ricktaylor:
Completely agree - I'm working on the parser updates right now that will allow this to build on top

sylvain-pierrot:
> Completely agree - I'm working on the parser updates right now that will allow this to build on top

Amazing! Look forward to moving forward

ricktaylor:
I have moved all the salient points into main, and the updates are queued up in the refactor/parser branch

ricktaylor:
This should close because 9 of its 12 commits are stale duplicates of feat/key-management (merging would re-delete design docs now on main), and the 3 unique commits implement a Dispatcher→BPA pipeline restructure that parallels — and conflicts with — the pipeline architecture settled in queue_architecture.md and bpa/docs/streaming_pipeline_design.md, including introducing a competing bundle_processing_design.md. The branch has been untouched since mid-May and is ~200 commits behind. If any part of the restructure addresses something the settled design doesn't, happy to see it re-proposed against current main. Branch will be deleted with the close.

ricktaylor (changes requested):
This PR is should be closed

PR #545, opened 2026-05-16
Feat/shared watcher

sylvain-pierrot:
## Description

> Depends on #543 

The file watcher logic is duplicated between `bpa-server` and `tvr`. Both implement the same pattern: debounced file monitoring with native/poll modes, parent directory watching, event filtering.

This PR extracts the shared watcher into `hardy-async` behind a `watcher` feature flag. Both crates now use `hardy_async::watcher::watch()` instead of their own implementations.

## What changed

**`hardy-async`**: new `watcher` module with `WatchMode` enum and `watch()` async function. Gated behind the `watcher` feature (depends on `notify`, `notify-debouncer-full`, `flume`, `serde`).

**`bpa-server`**: removed local `watcher.rs`, updated imports to `hardy_async::watcher`. Dependencies (`notify`, `notify-debouncer-full`) removed from `Cargo.toml`.

**`tvr`**: removed local `watcher.rs`, inlined the watch logic in `main.rs`. Config changed from `watch: bool` (default `true`) to `watch: Option<WatchMode>` (default absent/disabled). TVR now supports both native and poll modes.

### bpa-server (unchanged)

```yaml
bpsec:
  keys-file: /etc/hardy/keys.jwks
  watch: poll

static-routes:
  watch: native
```

### tvr (updated)

```yaml
# Before
watch: true

# After
watch: native   # or "poll" for Docker, omit to disable
```

## Shared watcher API

```rust
use hardy_async::watcher::{WatchMode, watch};

watch(&path, WatchMode::Poll, cancel_token, || async {
    // called on file create/modify/remove
}).await;
```

Monitors the file's parent directory (non-recursive), filters events to the target file, debounces with 1-second window. Runs until the cancellation token is triggered.


bpa-server/src/config/bpsec.rs
```diff
@@ -0,0 +1,346 @@
...
+#[derive(Clone, Serialize, Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub struct Config {
+    /// Path to a JWK Set file (RFC 7517 Section 5).
+    /// The file SHOULD have restrictive permissions (0600 on Unix).
+    pub keys_file: PathBuf,
+
+    /// Watch the key file for changes and reload automatically.
+    /// Values: "native" (inotify/kqueue), "poll" (works in Docker). Absent to disable.
+    #[serde(default)]
```
ricktaylor, line 21: Only criticism - can we make the default "native" and add an "off"?  We've changed the 'reasonable default' from "watching" to "not watching", and I think that's a regression

ricktaylor (changes requested):

bpa-server/src/config/bpsec.rs
sylvain-pierrot: After a double check on main, tvr is true by default and static-routes is false. 

What do you think about:

```
pub enum WatchMode {
    None,
    Poll,
    Native
}
```

With default on None or Native?

ricktaylor: That works - with default on Native - it's most likely what the user wants.  I think to disable it, or work-around a container issue they should configure something, otherwise it "just works"

sylvain-pierrot: I’ve made the change

ricktaylor (changes requested):
Can we split the BPSec key changes out to the other PR, and just have the base Watcher changes, and updates to existing uses of Watcher - then we can merge before the BPSec changes

PR #547, opened 2026-05-19
Feat/shared watcher standalone

sylvain-pierrot:
## Description

The file watcher logic is duplicated between `bpa-server` and `tvr`. This PR extracts it into `hardy-async` behind a `watcher` feature flag and reorganizes `bpa-server` static routes along the way.

## What changed

**`hardy-async`**: new `watcher` module with `watch()` async function. Monitors a file's parent directory, filters events to the target file, debounces with a 1-second window. Runs until a cancellation token is triggered.

**`bpa-server`**: split monolithic `static_routes/mod.rs` into `agent.rs` and `loader.rs`. Extracted static routes config into `config/static_routes.rs`. Uses shared watcher.

**`tvr`**: removed local `watcher.rs`. Uses shared watcher.

## Watch modes

Both `bpa-server` and `tvr` expose a `watch` config field with three values:

- **`native`** (default): uses OS events (inotify on Linux, kqueue on macOS). Fast, but does not work across Docker bind mounts.
- **`poll`**: periodically stats the file (~2s interval). Works everywhere including Docker bind mounts.
- **`none`**: disables file watching entirely.

```yaml
static-routes:
  watch: poll    # or "native" (default), "none" to disable
```


ricktaylor (dismissed):
Looks AMAZING!

ricktaylor (approved):

PR #548, opened 2026-05-20
crc: consolidate CRC emit/verify into a single Digest type

ricktaylor:
## Summary

- A a single `crc::Digest` that owns the per-CRC-type facts (algorithm, value length, CBOR byte-string head byte, big-endian finalisation). Adding a new CRC type now means one match arm to update, not three.
- `append_crc_value` and `parse_crc_value` survive as thin wrappers around `Digest` for ergonomic Vec-in / Vec-out use; new call sites (e.g. the push-parser in `bundle/raw_parse.rs`) can use `Digest` directly.
- Drop `Error::InvalidLength` — wrong-length CRC bytestrings now surface as `IncorrectCrc`. Same condition, simpler error surface.

### DIFF
--- a/bpv7/src/crc.rs
+++ b/bpv7/src/crc.rs
@@ -18,10 +18,6 @@ pub enum Error {
     #[error("Invalid CRC Type {0}")]
     InvalidType(u64),
 
-    /// Indicates that the CRC value in a block has an unexpected length.
-    #[error("Block has unexpected CRC value length {0}")]
-    InvalidLength(usize),
-
     /// Indicates that a block has a CRC value but no CRC type was specified.
     #[error("Block has a CRC value with no CRC type specified")]
     UnexpectedCrcValue,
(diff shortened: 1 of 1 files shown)


cbor/src/tests/mod.rs
sylvain-pierrot, line 1: Quick comment, why are the tests in `src/`?

sylvain-pierrot (approved):

cbor/src/tests/mod.rs
ricktaylor: I'm slowly moving tests out of src as I fidn them - not going to make the cange in this PR as it is so small

bpv7/src/crc.rs
```diff
@@ -196,29 +165,99 @@ pub(super) fn parse_crc_value(
...
+
+    /// Finalises the digest and returns the computed CRC value as
+    /// big-endian bytes ([`value_len`](Self::value_len) bytes long).
+    /// Callers comparing against a wire-form value bytestring are
+    /// responsible for the length pre-check (compare against
+    /// [`value_len`](Self::value_len)) before comparing the bytes.
+    pub fn finalize(self) -> Vec<u8> {
+        match self.state {
+            DigestState::Crc16(d) => d.finalize().to_be_bytes().to_vec(),
+            DigestState::Crc32(d) => d.finalize().to_be_bytes().to_vec(),
```
sylvain-pierrot, line 271: All the PR is very clean. I would raise only one comment. This is micro-optimization. 🙂 

If we care about zero-alloc parsing,  `finalize()` allocates a `Vec<u8>` (2 or 4 bytes) on every CRC check via `.to_be_bytes().to_vec()`. Since the parse path only compares the result against a slice, you could add a `verify(&self, expected: &[u8]) -> bool` method that compares internally without allocating. The emit path can keep `finalize()` as-is since it writes into a Vec anyway.

sylvain-pierrot: The change would be adding this method to Digest:                                                                                     
  ```rust                                                                                                                                      
  pub fn verify(self, expected: &[u8]) -> bool {                                                                                        
      match self.state {                                                                                                                
          DigestState::Crc16(d) => expected == d.finalize().to_be_bytes(),                                                              
          DigestState::Crc32(d) => expected == d.finalize().to_be_bytes(),                                                              
      }                                                                                                                                 
  }                                                                                                                                     
  ```                                                                                                                                   
  Then parse_crc_value changes from:                                                                                                    
  ```rust 
  if digest.finalize() != data[crc.start..crc.end] {                                                                                    
      return Err(Error::IncorrectCrc);                                                                                                  
  }                                                                                                                                     
  ```                                                                                                                                      
  to:                                                                                                                                   
 ```rust                                                                                                                                       
  if !digest.verify(&data[crc.start..crc.end]) {
      return Err(Error::IncorrectCrc);
  }                                                                                                                                     
  ``` 
The `.to_be_bytes()` returns [u8; 2] or [u8; 4] on the stack. Comparing directly against that avoids the `.to_vec()` heap allocation. `finalize()` stays unchanged for the emit path.

sylvain-pierrot (changes requested):

bpv7/src/crc.rs
ricktaylor: Done - good catch!

sylvain-pierrot (approved):
LGTM!

PR #550, opened 2026-05-20
refactor: replace RoutingSink with channel-based RoutingContext

sylvain-pierrot:


ricktaylor:
This should close because the specific Context-splitting approach here doesn't fit the architecture that has since settled on main, although the channel-based Context concept itself pointed in the right direction. The streaming pipeline design (bpa/docs/streaming_pipeline_design.md) now defines how component/BPA communication is shaped, and the Sink lifecycle rework planned in TODO.md takes a different route (component-owned Arc<dyn Sink> + registry Weak + reconciler task). Closing as a useful but dead-end experiment — the lessons fed the current design. Branch will be deleted with the close.

ricktaylor (changes requested):
This PR should be closed

ricktaylor (changes requested):
Should be closed

PR #551, opened 2026-05-22
refactor: decouple Sink lifecycle from CLA registry with drop reconciler

sylvain-pierrot:
## Problem

When a CLA's `Sink` is dropped, the `Drop` impl spawns an async task via `hardy_async::spawn!` to call `registry.unregister()`. If the tokio runtime is shutting down (or already gone), the spawn fails silently. The CLA entry is never removed, which means:
                                                                                                                                                  
  - Peers remain in the `PeerTable` with active egress queues, holding `Arc`s to the store and dispatcher
  - RIB routes stay installed pointing at a dead CLA
  - Metrics (`bpa.cla.registered`, `bpa.fib.entries`) never decrement                                                                             
  - The `address_type` mapping stays in the RIB, routing inbound bundles to a CLA that no longer exists
                                                                                                                                                  
This matters for long-running BPA instances where CLAs connect and disconnect over time: each failed cleanup is a stale route and a slow resource leak.

## Solution

A dedicated reconciler task is spawned once at registry construction. `ClaSink::drop()` pushes a `Weak<ClaEntry>` onto an unbounded `flume` channel (sync, never blocks, never fails). The reconciler receives it and calls `registry.unregister()` while the runtime is still alive. If the entry was already explicitly unregistered, the reconciler finds nothing to do (idempotent via `HashMap::remove`).


ricktaylor (changes requested):
This should close because its core change — moving off flume onto the hardy-async channel abstraction — landed on main independently via the refactor/channels work (169541d7, 5801953a, c619cd15), and the Sink lifecycle rework is now planned along different lines (see TODO.md). Anything here beyond the channel migration that still applies is worth re-proposing as a focused PR against current main.

Issue #523, opened 2026-04-30
Pre-built compose.yml using published OCI images


ricktaylor:
## Problem

The current `compose.yml` uses `build:` directives that compile Hardy from source:

```yaml
hardy:
  build:
    context: .
    dockerfile: ./bpa-server/Dockerfile
    args:
      CARGO_ARGS: "--features ..."
```

This requires a full Rust toolchain and takes several minutes. Users who just want to run Hardy should be able to docker compose up with the published ghcr.io images.

### 1. Replace build: with published images

The hardy and hardy-debug services should reference the multi-arch images already published by the oci-images.yml workflow:

hardy:
  image: ghcr.io/<org>/hardy-bpa-server:latest

### 2. Config file distribution

The current compose mounts ./bpa-server/config.yaml from the source tree:

volumes:
  - ./bpa-server/config.yaml:/etc/hardy/config.yaml:ro

For users without a checkout, we need to decide how to provide configuration:

Options:
- A) Ship a standalone config.yaml alongside the compose file (e.g., in a deploy/ directory or GitHub release artifact)
- B) Embed a default config in the container image and only mount overrides
- C) Drive all configuration via environment variables (already partially supported via config crate env var mapping)
- D) Provide a docker compose profile that generates a default config on first run

### 3. Scope

- Should the pre-built compose replace the existing compose.yml, or live alongside it (e.g., compose.yml for development, deploy/compose.yml for users)?
- Should TCPCLv4 and TVR services be included in the compose?
- Should the hardy-debug profile also use published images?

### Current State

- 4 OCI images published to ghcr.io (bpa-server, tcpclv4-server, tvr, tools) with multi-arch support (amd64 + arm64)
- Config is driven by config crate with TOML file + env var overlay + CLI args
- compose.yml currently serves dual purpose: CI storage tests and developer workflow

PR #557, opened 2026-05-27
feat(docker-compose): use public hardy-bpa-server image url

sylvain-pierrot:
#523

Use the ghcr's bpa-server image into the docker compose. `build` and `image` can live together in a **compose service**, and so, running the compose command (`docker compose up`) wil pull the bpa-server image from ghcr, otherwise you can use `—build` option for building an image from local project.


ricktaylor (approved):
LGTM

PR #558, opened 2026-06-01
Feat/service file

sylvain-pierrot:


ricktaylor (changes requested):
## Review: `feat/service-file`

  Nice addition — a clean, well-documented [`hardy-file-service`](https://github.com/ricktaylor/hardy/tree/feat/service-file/file-service) crate that bridges a filesystem outbox/inbox to the BPA over gRPC as an Application endpoint. It follows the established
  [`tcpclv4-server`](https://github.com/ricktaylor/hardy/tree/feat/service-file/tcpclv4-server) pattern closely (config layering, `listen_for_cancel`, sink-drop unregistration).

  I confirmed the outbox concurrency model is correct: `select!`-ing on the `BoundedTaskPool::spawn` future in [`outbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs) awaits the *semaphore permit* (backpressure), then drops
  the `JoinHandle` so the send runs detached-but-tracked, with `senders.shutdown()` draining it. 👍

  A few issues below — I'd consider **1**, **2**, and **7** blocking.

  ### Correctness

  **1. Quarantined files can't be resubmitted via the documented procedure.**
  [`process_file`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs) renames `name` → `name.processing` *first*, then on failure `move_to_errors` lands it in `errors/` as `name.processing`. But
  [`is_processable`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs#L118) rejects the `.processing` extension, so following the [`usage.md`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/docs/usage.md)
  recovery step (`mv outbox/errors/failed_file.bin outbox/`) silently fails — the real filename is `failed_file.bin.processing`. Fix: strip the `.processing` suffix in `move_to_errors` (also makes `errors/` human-readable).

  **2. No startup scan / crash recovery.**
  The watcher in [`outbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs) only sees events fired *after* `watch()`. So (a) files already in the outbox at startup are never sent until touched again, and (b) a crash between the
  `.processing` rename and send completion strands the payload forever (orphaned `*.processing`, original gone). A startup directory scan that processes plain files and recovers stale `*.processing` files would close both gaps. Currently called out only as a
  limitation in [`design.md`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/docs/design.md).

  ### Robustness / shutdown

  **3. [`FileService::unregister()`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/lib.rs) doesn't drain its own task pool.**
  It only calls `sink.unregister()` and relies on the BPA round-tripping `on_close` → `on_unregister` → `self.tasks.shutdown()`. In [`main.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/main.rs), after `service.unregister().await`
  returns we go straight to exit, so in-flight outbox sends may be cut off. An explicit `self.tasks.shutdown().await` in `unregister()` makes this deterministic.

  **4. `process_file` ignores cancellation.**
  If `sink.send` hangs (unresponsive BPA), `senders.shutdown().await` in [`outbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs) waits on it forever and graceful shutdown hangs. Consider a `select!` against the cancel token 
  around the send.

  ### Data-loss surfaces (partly inherent — flagging)

  **5. Inbox write failure is silent loss.**
  [`on_receive`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/lib.rs) has no way to signal failure, so the BPA treats the bundle as delivered even if the 
  [`inbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/inbox.rs) write fails. Documented in `usage.md` as "monitor logs," which is acceptable given the trait — just noting it.

  **6. Empty files are silently deleted** after being claimed in [`process_file`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs). A zero-length payload is a legitimate thing to send; at minimum document this in 
  [`design.md`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/docs/design.md).

  ### Tests

  **7. No tests in the crate.**
  This is a test-heavy project (per-crate coverage reports, TRR-approved plans). `is_processable` / `is_file_ready` in [`outbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs), the filename generation in 
  [`inbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/inbox.rs), and config loading in [`config.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/config.rs) are all trivially unit-testable. The 
  [`tcpclv4-server` config tests](https://github.com/ricktaylor/hardy/blob/feat/service-file/tcpclv4-server/src/config.rs) can be mirrored almost verbatim.

  ### Minor / polish

  - **Doc inconsistency:** [`config.yaml`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/config.yaml) uses `destination: "ipn:1.42"`; [`usage.md`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/docs/usage.md) and the 
  [`main.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/main.rs) CLI help use `"ipn:2.1.42"`. Pick one.
  - **Possibly unused dep:** `humantime` is a direct dependency in [`Cargo.toml`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/Cargo.toml) but only `humantime-serde` appears used.
  - **"Failed to claim" logged at `error!`** in [`outbox.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/outbox.rs) for a usually-benign duplicate-event race — consider `debug!`.
  - **`default_config_path` isn't platform-gated** like `tcpclv4`'s in [`config.rs`](https://github.com/ricktaylor/hardy/blob/feat/service-file/file-service/src/config.rs). Defensible (inotify ⇒ Linux-only) — just flag it as deliberate.
  - **Naming proximity** with the existing [`hardy-file-cla`](https://github.com/ricktaylor/hardy/tree/feat/service-file/file-cla) (a convergence layer) could confuse; a one-line cross-reference in `design.md` would help.

sylvain-pierrot:
Thank you for your review. I’ve addressed all your comments 🙂

ricktaylor:
Thanks for the thorough rework, Sylvain. **All seven points from the last round are properly addressed**, including the three blockers:

  - ✅ Quarantined files keep their original name (`errors/foo.bin`, not `.processing`), so the documented `mv errors/foo.bin outbox/` resubmit works.
  - ✅ Startup recovery (`emit_existing_files`) handles pre-existing files and stranded `*.processing` files, and starting the watch *before* the scan neatly closes the race.
  - ✅ `unregister()` drains its pool; `process_file` selects against the cancel token; comprehensive tests (~40).
  - ✅ Minors too — `humantime` dropped, "Failed to claim" → `debug!`, the `file-cla` cross-reference, EID standardization.

  I also re-confirmed the outbox concurrency model, and empirically checked notify 8.2's rename events (it emits a standalone `Modify(Name(To))` alongside the coalesced `Both`), so keying `is_file_ready` on `To` is correct.

  The recovery/cancel/shutdown machinery is all new since the last pass, so a few things it introduced. **Two I'd ask you to fix before merge; the rest are polish.**

  ### 🔴 Should-fix

  **1. Clean shutdown dead-letters in-flight sends, which are then never auto-retried.** `outbox.rs:247–267`, `lib.rs:67–70`
  A send still in flight at shutdown takes the `cancel.cancelled()` branch → `move_to_errors` → `errors/`. And because `unregister()` calls `sink.unregister()` *before* `self.tasks.shutdown()`, racing sends also fail with `Disconnected` and land in `errors/` via the
  `Err` arm. Since startup recovery only rescues `*.processing` (it deliberately skips `errors/`), a payload interrupted by a routine restart is stranded until someone moves it back manually. This contradicts `design.md:110–113` ("waits for all in-flight send tasks to
  complete"), and for the large-file / slow-BPA case the docs target, the window is real.
  *Suggested fix:* on the **cancellation** branch, rename `.processing` back to the plain name so the next startup resends it, and reserve `errors/` for genuine send failures. Also drain `self.tasks` *before* `sink.unregister()` so in-flight sends finish cleanly
  instead of hitting `Disconnected`. Update the design-doc shutdown paragraph to match whatever we land on.

  **2. The inotify-only watcher silently misses live writes on Docker bind mounts — and we already have the fix in-tree.** `outbox.rs:31–54`
  `usage.md` leads with bind-mounted volumes, but `RecommendedWatcher` is pure inotify, which doesn't receive events for host-bind-mounted dirs under Docker Desktop (macOS/Windows) or some overlay setups. The startup scan catches what's present at boot, but later drops
  are invisible. `hardy_async::watcher` (`async/src/watcher.rs`) already solves this with `WatchMode::Poll` ("Works everywhere including Docker bind mounts, ~2s latency"), built on `notify_debouncer_full` like `file-cla`. Can we reuse it, or add a poll-fallback config
  knob? (Keep your `CLOSE_WRITE`/`MOVED_TO` model — it's better than `file-cla`'s `Create`-based one, which can read half-written files. This is only about the poll capability.)
  Related: `design.md:131` says it "falls back to polling on non-Linux" — `RecommendedWatcher` doesn't do that; it uses native backends, and `CLOSE_WRITE` is Linux-only, so non-Linux is effectively unsupported (which the "Linux only" limitation already states — the
  polling line contradicts it).

  ### 🟡 Nice-to-have

  **3. Re-registration would start a second watcher on a stale sink.** `lib.rs:79–92`
  `on_register` stores the sink via `Once` (keeps the first) but unconditionally calls `outbox::run` each time, so a second registration spawns a second watcher bound to the dead sink. One registration per process makes this unlikely — a guard or `debug_assert` would
  harden it.

  ### 🟢 Polish

  - [ ] `outbox.rs:188` — `is_processable` uses `parent().ends_with("errors")`, so configuring `outbox = /…/errors` makes *every* file unprocessable. Niche footgun.
  - [ ] No `config.rs` tests — outbox/inbox coverage is great, but env/file layering and humantime parsing are untested. The tcpclv4-server config tests port over almost verbatim.
  - [ ] `config.rs:91` — the "Loaded configuration from" message checks `config_file.exists()` on the extensionless default path, so it's suppressed exactly when a `.yaml/.toml/.json` *was* auto-discovered.
  - [ ] `usage.md:64–67` — inbox filename example still shows `ipn_2.1.42_…`; everything else is now `ipn:1.42`.

  Overall in good shape — happy to re-review once #1 and #2 are sorted. Thanks again!

ricktaylor (changes requested):
Thanks for the rework, Sylvain. Re-reviewed the three new commits on top of the rebase — builds clean, clippy clean, all 53 tests green. The last round is essentially sorted:

✅ **#3 re-registration guard** — `on_register` early-returns on an existing sink (lib.rs:77).
✅ **Shutdown (#1, code side)** — `unregister()` drains before unregistering the sink (lib.rs:67–69), cancel branch restores `.processing` → plain name instead of dead-lettering (outbox.rs:246–251). Routine restarts no longer strand payloads in `errors/`. 👍
✅ **`is_processable` errors-dir footgun** — gone; `errors` is its own configurable dir.
✅ **config tests** + **usage.md filename/paths polish**.

On the inotify-vs-poll watcher (#2 last round): you're right, and I'll withdraw it. `PollWatcher` can only emit Create/Modify/Remove — it can't produce `Close(Write)` or `Modify(Name(To))`, so poll would silently process nothing under the current readiness model, and
the only way to make it work is a Create+debounce-settle heuristic that reintroduces the half-written-file race you deliberately designed out. Inotify-only is the right call. Two small asks instead:

🔴 **Should-fix**

**1. Shutdown design-doc paragraph contradicts the code. design.md:113–117**
The code's right, but the paragraph claims both "waits for all in-flight send tasks to complete" / "ensuring in-flight sends complete against a live connection" *and* "cancelled and renamed back." `tasks.shutdown()` cancels the shared token that `process_file`
selects on, so an in-flight send takes the cancel branch and rolls back — it doesn't complete. Please reword to rollback-and-resume-on-restart and drop the "complete against a live connection" clause. (True drain-to-completion would need the event channel closed + a
separate token for in-flight sends — not in scope.) One honest caveat for Limitations while you're there: if `sink.send` succeeds on the BPA but the cancel branch wins the race, the file is restored and resent next startup → duplicate bundle. Fine for at-least-once
DTN, just worth stating.

🟢 **Polish**

- **usage.md `docker run` example is broken** — it mounts `./outbox:/data/outbox` but sets neither `HARDY_FILE_SERVICE_OUTBOX` nor `-c`, so the service watches the default `/tmp/hardy/outbox` and never sees the drops. The compose example below is correct. Please
align them (and since the watcher is inotify-only, maybe lead the Docker section with native-Linux + the Docker Desktop caveat rather than bind mounts generally).
- **A one-line comment at the watcher (outbox.rs:27)** noting the inotify-only / CLOSE_WRITE+MOVED_TO choice is intentional (poll can't deliver those kinds) — saves the next reviewer from suggesting poll.
- **config.rs** — "Loaded configuration from…" is now gated on `required`, so it's suppressed exactly when a config is auto-discovered at the default path. Minor/harmless.

Down to the shutdown paragraph and a couple of doc nits — happy to approve once those land. Thanks again!

ricktaylor:
You could use `select_biased` to avoid the send vs cancel race...

ricktaylor (changes requested):
🟢 Nit (config.rs:115) — the "Loaded configuration from…" message is gated on `required`, which is only true for an explicit `-c`/`HARDY_FILE_SERVICE_CONFIG_FILE` path. So when a config is auto-discovered at the default `/etc/hardy/file-service[.yaml|.toml|.json]`,
it loads silently — the operator gets no confirmation in exactly the deployment case where it's most useful.

(For context, this was the over-correction of the earlier `|| config_file.exists()` version: `exists()` was unreliable because `config_file` is the extension-less base path that `config::File::with_name` probes.)

Trivial fix — log the resolved path unconditionally:

    eprintln!("Loaded configuration from '{}'", config_file.display());

…or drop the message entirely if it's not pulling its weight. Either's fine, non-blocking.

Note I'm suggesting unconditional logging because tracing isn't initialized yet at config-load time (that's why it's eprintln!, and why you can't just demote it to debug!). Unconditional eprintln! is the genuinely 2-second fix.

ricktaylor (approved):
LGTM!

PR #561, opened 2026-06-11
Feat/transactional routing table

sylvain-pierrot:
# Transactional Route Table

## What is the Route Table?

The route table is the core data structure that decides where bundles go. When a bundle arrives, the BPA looks up its destination in the route table to find the next hop. This lookup happens on every single bundle, making it the hottest path in the system.

Routes come from two sources:

- **Routing agents** (external): operators configure static routes, or protocols like TVR compute time-varying routes. These can be misconfigured.
- **BPA internals**: CLA peers connecting/disconnecting, services registering/unregistering. These are always correct.

### A typo could take down all routing

When the static routes file is watched for changes, a syntax error during reload caused all existing routes to be deleted. The parser returned an empty list on error, and the reload logic interpreted "empty" as "remove everything". One bad edit and the node stops forwarding.

### Misconfigured routes could be partially applied

Routes were added one at a time. If an operator submitted 10 routes and the 5th one was invalid, the first 4 were already in the table. No way to roll back. Worse, some misconfigurations (like A routes to B, B routes to C, C routes back to A) only become visible when you look at all routes together.

### Route lookups were slower than necessary

Every bundle lookup had to coordinate with potential route updates through a read-write lock. Even though lookups happen thousands of times per second and route changes happen a few times per hour, every lookup paid the synchronization cost.

### Transactional updates

Routes are now updated as a transaction. Instead of adding and removing routes one by one, agents submit a batch: "add these routes, remove those routes". The RIB validates the entire resulting table before applying anything. If any route is invalid, the whole batch is rejected and the table stays unchanged. This is the same principle as a database transaction: all or nothing.

### Lock-free lookups

The table uses an atomic pointer swap pattern (ArcSwap). Readers load a snapshot of the table with a single atomic operation and work on it without any lock. Writers build a new table, validate it, then swap the pointer. Readers and writers never interfere with each other.

### Type-safe action split

The type system prevents mixing up route types. Routes from agents (`Via`, `Reflect`, `Drop`) go through validation. Routes from BPA internals (`Forward` to a CLA peer, `Local` service delivery) skip validation because they are always correct by construction. Two different types enforce this at compile time: you cannot accidentally insert agent-configured routes through the internal path or vice versa.

## What gets validated?

When an agent submits routes, each route is checked:

- A `Via` route cannot point to a null endpoint
- A `Via` route cannot point to a destination that matches its own pattern (would loop immediately)
- A `Via` route cannot point to the local node (would route to ourselves)
- A `Reflect` route cannot match the local node's own address

After individual checks pass, the full table is checked for transitive loops: if route A points to B, B points to C, and C points back to A, the commit is rejected.

## What happens on failure?

If validation fails, the error is returned to the agent. The route table is completely unchanged. The agent can log the error, retry with different routes, or keep running with the previous configuration.

For static route file reloads specifically: if the file has parse errors, the reload is skipped entirely and the previous routes stay in place. No more accidental wipeout.

## Module structure

```
bpa/src/routing/
  mod.rs           - public API: RoutingAgent, RoutingSink, Error, re-exports
  route.rs         - Route struct with via/drop/reflect constructors
  rib.rs           - Rib: find, agent lifecycle, route operations
  sink.rs          - RoutingSink implementation
  static_agent.rs  - StaticRoutingAgent helper
  table/
    mod.rs         - shared types: Entries, Error, FindResult
    action.rs      - RouteAction, InternalAction, Action
    atomic.rs      - AtomicRouteTable (ArcSwap + write mutex)
    virtual.rs     - VirtualRouteTable (validation + commit)
```

## Breaking changes

- `hardy_bpa::routes` renamed to `hardy_bpa::routing`
- `routes::Action` renamed to `RouteAction`
- `RoutingSink::add_route`/`remove_route` replaced by `update_routes(add, remove)`
- `StaticRoutingAgent::new` takes `&[Route]` instead of `&[(EidPattern, Action, u32)]`


ricktaylor (changes requested):
Sent to change details over Slack

PR #562, opened 2026-06-12
feat: BPSec key configuration with EID-pattern bindings and role-gate

ricktaylor:
## Summary
  
  Adds BPSec key configuration to `bpa-server`: keys are loaded from a JWKS file (RFC 7517) and bound to security source EIDs via EID patterns, with a verifier/acceptor/source role per binding and automatic hot reload when the key file changes.

  This is a focused harvest of the key-loading work from `feat/key-management` (thanks @sylvain — most of the config loading, pattern source, and watcher wiring originates there), reworked to avoid the wider pipeline refactors on that branch:
  the existing `KeyProvider` trait is kept, and all policy lives in the server binary rather than the BPA library.

  ## Design

  **Roles are expressed through key release, not a policy enum.** `KeySource::key(source_eid, operations)` returning a key is a policy claim — failure with a released key indicates corruption (RFC 9172 §5.1.1 strip/fatal). Withholding a key
  yields `NoKey`, and the block is forwarded intact for a downstream security acceptor. So a *verifier* binding releases keys only for `verify` (BCBs ride through encrypted), an *acceptor* additionally releases `decrypt`/`unwrapKey`, and a
  *source* binding only releases protection keys (inert until an egress signing path exists). No role enum reaches bpv7, and the mechanism matches the direction of the parse refactoring work, so this should merge cleanly with it later.
  
  **Policy lives in `bpa-server`** (mirroring the static_routes split): `bpsec.rs` holds `SecurityRole`, `PatternKeySource` (specificity-ordered EID-pattern bindings → kids, role-gated lookup, then `key_ops` matching), `PatternKeyProvider` (an
  `ArcSwap` over the source, so reload is lock-free), and `watch_keys()` (file watcher task; a failed reload logs and keeps the previous keys).

  **The bpa library change is minimal:** `BpaBuilder::key_provider()` sets a single creation-time provider. The WIP multi-provider `Registry` is removed — its first-key-wins aggregation over an unordered map made policy resolution
  non-deterministic, and the per-bundle lock/allocation fan-out is gone with it. An unconfigured node gets a null provider serving `KeySet::EMPTY`, i.e. every BPSec lookup is `NoKey`, preserving existing behaviour.

  ## Configuration
  
  ```yaml
  bpsec:
    keys-file: /etc/hardy/keys.jwks   # JWK Set, should be chmod 0600
    watch: poll                       # native (default) | poll | none
    bindings:
      - match: "ipn:*.*"
        role: acceptor                # verifier (default) | acceptor | source
        keys: ["test-hmac", "test-aes"]
```

  Key validation at load: symmetric (oct) non-empty keys only, key_ops required, duplicate kid rejected, bindings must reference known kids. Group/other file permissions produce a warning.

  Testing
  
  - Unit/integration: pattern matching and specificity ordering, role gating (including most-specific-role-wins), key_ops routing, provider hot-swap snapshot isolation, JWKS validation failures, config parsing.
  - RFC 9173 Appendix A.2 vector: a payload BCB rides through parse intact for both roles; at delivery a verifier gets NoKey while an acceptor recovers the plaintext.
  - Manual: hot reload verified live — valid key-file edit reloads, corrupt file keeps previous keys, clean shutdown.
  - cargo test -p hardy-bpa -p hardy-bpa-server, workspace clippy and rustfmt all clean.

  Out of scope (deliberately)

  - BIB removal at acceptor and destination-gated acceptor semantics — these belong to the full security policy layer coming with the parse refactor.
  - Egress signing (which would activate source role bindings).
  - Asymmetric keys (config rejects non-oct keys for now).

### CODE REVIEW

File: bpa-server/src/config/mod.rs [hunk 1]
```diff
@@ -1,17 +1,13 @@
...
-use hardy_async::available_parallelism;
-use hardy_async::watcher::WatchMode;
-use hardy_bpa::filter::rfc9171;
-use hardy_bpa::node_ids::NodeIds;
+use hardy_async::{available_parallelism, watcher::WatchMode};
+use hardy_bpa::{filter::rfc9171, node_ids::NodeIds};
 use hardy_bpv7::eid::Service;
 use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, path::PathBuf};
 use tracing::Level;
```
sylvain-pierrot: Imports ordering?

- std/core

- ext. crates

- super/crate

bpa-server/src/bpsec.rs
```diff
@@ -0,0 +1,664 @@
...
+
+impl Default for PatternKeyProvider {
+    fn default() -> Self {
+        Self::new(PatternKeySource::empty())
+    }
+}
+
+impl hardy_bpa::keys::KeyProvider for PatternKeyProvider {
+    fn key_source(&self, _bundle: &hardy_bpv7::bundle::Bundle, _data: &[u8]) -> Box<dyn KeySource> {
+        Box::new(CurrentKeys(self.source.load_full()))
```
sylvain-pierrot, line 182: Arc wrapped with a Box?

bpa-server/src/config/bpsec.rs
sylvain-pierrot, line 1: I would suggest moving out from `config/` all the bpsec implementation not related to the configuration

sylvain-pierrot (changes requested):
The code looks good! I'd like to suggest a small reorganization though: the config folder in bpa-server is meant strictly for configuration handling, so I'd move any non-bspec configuration code out of there and into a dedicated file at the project root

ricktaylor:
Agreed, done. Moved the BPSec implementation — JWKS loading, key/binding validation, and the 0600 permission check — out of config/ into bpsec.rs, so config/ now holds only the serde structs. Specifics in the config/bpsec.rs thread.

bpa-server/src/config/bpsec.rs
ricktaylor: Done. The loading/validation/permission-check logic (and its tests) now lives in bpsec.rs as PatternKeySource::from_config; config/bpsec.rs is just the Config + KeyBindingConfig serde structs, mirroring config/static_routes.rs. main.rs and watch_keys call PatternKeySource::from_config directly.

bpa-server/src/config/mod.rs
ricktaylor: I'd rather keep these as a single contiguous block — it's cargo fmt --check clean and the form I'd like to converge the crate on. I know the older files (cla.rs, policy.rs, storage.rs, static_routes.rs) group std/ext/crate with blank lines; I'd sooner align those to the single-block form in a separate pass than split this one back out.

bpa-server/src/bpsec.rs
ricktaylor: The Box is fixed by KeyProvider::key_source's return type. The Arc is deliberate: key_source runs on the per-bundle parse path and KeySource::key hands back &Key, so the returned source has to own its keys — sharing the current table via Arc keeps that O(1) and gives each bundle a snapshot that's stable across a hot-reload, versus deep-cloning the whole key table on every bundle. CurrentKeys is just the newtype needed to impl KeySource for the Arc (orphan rule). I've added a doc comment capturing this. The only way to drop the Box-around-Arc entirely is to return Arc<dyn KeySource>, which also means changing bpv7's parse closures

bpa-server/src/config/mod.rs
sylvain-pierrot: I’d rather keep the std/ext/crate, I’ve been always using this organization.

bpa-server/src/bpsec.rs
sylvain-pierrot: ok, we’ll see later for that. Let’s keep as is for now.

bpa-server/src/config/bpsec.rs
sylvain-pierrot: perfect!

ricktaylor:
Close #297 

Update on #543

bpa-server/src/bpsec.rs
```diff
@@ -0,0 +1,934 @@
...
+            bindings: Vec::new(),
+        }
+    }
+
+    /// Loads keys from the JWKS file named in `config` and builds a
+    /// `PatternKeySource`.
+    ///
+    /// Every key must be a non-empty symmetric key (`kty: oct`) carrying a
+    /// `key_ops` field, and every binding must reference a known key id.
+    pub fn from_config(config: &crate::config::bpsec::Config) -> anyhow::Result<Self> {
```
sylvain-pierrot: This should live in the `config/`

bpa-server/src/config/mod.rs
ricktaylor: Done!

bpa-server/src/bpsec.rs
ricktaylor: Not so sure - this kind of spans the gap between Config structs and real impl - so I think it makes more sense to have it with the rest of the implementation

sylvain-pierrot: This is configuration validation..

bpa-server/src/main.rs
```diff
@@ -55,13 +55,29 @@ async fn main() -> anyhow::Result<()> {
...
-    let bpa = build(config, args.upgrade_storage).await?;
+    let bpsec_config = config.bpsec.take();
+    let key_provider = bpsec_config
+        .as_ref()
+        .map(|c| {
+            c.build()
+                .context("Failed to load BPSec configuration")
+                .map(|source| Arc::new(PatternKeyProvider::new(source)))
+        })
+        .transpose()?;
```
sylvain-pierrot: I’d move this in `build()` instead of passing it as a parameter

bpa-server/src/main.rs
```diff
@@ -129,29 +153,7 @@ async fn build(config: config::Config, upgrade_storage: bool) -> anyhow::Result<
...
-                routes_file
-            }
-        };
-
-        let agent = Arc::new(StaticRoutesAgent::new(
-            routes_file,
-            sr_config.priority,
-            sr_config.watch.into(),
-        ));
-
```
sylvain-pierrot, line 154: Good catch!

bpa/src/keys.rs
sylvain-pierrot, line 1: NIce, very well-defined

bpa-server/src/main.rs
ricktaylor: Good call!

bpa-server/src/bpsec.rs
ricktaylor: Moved

bpa-server/src/main.rs
ricktaylor: Done — build() now constructs and registers the PatternKeyProvider from the bpsec config and returns it, so main no longer builds it or passes it in. It still has to return the provider (and main retains the bpsec config) because watch_keys needs both for hot-reload, which is why the construction was hoisted out originally — but the provider is built in build() now, and the config goes in as a borrow rather than a pre-built object.

bpa-server/src/main.rs
```diff
@@ -109,6 +120,18 @@ async fn build(config: config::Config, upgrade_storage: bool) -> anyhow::Result<
         builder = builder.service_priority(service_priority);
     }
 
+    let key_provider = bpsec_config
```
sylvain-pierrot: Why do you need a map here? 
We cannot do something like ?
```
 if let Some(bpsec_config) = bpsec_config {
        let source = bpsec_config
            .build()
            .context("Failed to load BPSec configuration")?;
        builder = builder.key_provider(KeyProvider::new(source));
    }
```

bpa-server/src/main.rs
```diff
@@ -55,13 +55,20 @@ async fn main() -> anyhow::Result<()> {
...
 
     let tasks = TaskPool::new();
     hardy_async::signal::listen_for_cancel(&tasks);
 
+    if let Some(bpsec_config) = bpsec_config
+        && let Some(provider) = key_provider
+    {
+        bpsec::watch_keys(&tasks, bpsec_config, provider);
+    }
+
```
sylvain-pierrot: What do you think about adding a method on the bpa to swap the key source? This would avoid the hack to get the `key_provider` from `build()`, and you could pass the bpa as parameter to `bpsec::watch_keys`

ricktaylor: I'm really not sure swapping providers makes any operational sense - I think it opens up all kinds of security/operational issues.  I think as an operator, you pick your provider when you run an instance.  If you need to do something as radical as change the provider of crypto keys, you really should be forced to restart the process.

In short - I don't think something as important of a crypto-key provider is something that can get programmatically hot-swapped

bpa-server/src/main.rs
ricktaylor: Good call — the map/transpose was only there to thread ? out of the closure, which isn't needed. Switched to a plain if let ... { ... ? ... }, matching the static_routes block just below. The one difference from your snippet is the let mut key_provider = None; up front: build() has to return the provider so main() can wire the hot-reload watcher, so I capture it inside the block rather than just dropping it on the builder.

bpa-server/src/main.rs
sylvain-pierrot: Not swapping the key provider, but the key source, based on a setter method living in the bpa

sylvain-pierrot: Basically having access internally to the key provider from the bpa

ricktaylor: I'd rather not add a key-source setter to Bpa. The BPA holds Arc<dyn KeyProvider>, and key_source() is per-bundle by design (the provider = flow selector / source = per-block resolver split) — there's no single "the key source" at that layer to set. PatternKeyProvider only looks like it has one because it hands out a global ArcSwap source; making Bpa aware of that would mean either a set_source(Box<dyn KeySource>) on the trait (heterogeneous → downcasting, plus a no-op default for NullKeyProvider) or swapping the whole provider, which we agreed against.

It's also the wrong layer: reloading a JWKS file is a bpa-server/operator concern, and bpa deliberately stays unaware of it — same as storage/CLAs/routing agents, which are all construction-time with no runtime set_x on Bpa. Key rotation is encapsulated in PatternKeyProvider's ArcSwap, invisible to the core, which is where I think it should stay. Returning the provider handle from build() is just the usual "keep a handle to what you registered so you can manage it" pattern — the handle is the reload surface for the watcher.

sylvain-pierrot: I see your point, fair enough. I would be happy if we could eliminate this tuple from `build()`

sylvain-pierrot: Otherwise, I don’t have any other comment for your PR!

sylvain-pierrot (dismissed):
Just one non-blocking comment left. If we cannot figure it out, I’m happy with your current PR 
LGMT!

ricktaylor:
Removed the tuple

sylvain-pierrot (approved):

PR #563, opened 2026-06-12
fix(tcpclv4): unify registry lock to prevent peer dedup race

sylvain-pierrot:


ricktaylor (changes requested):
Sent you a long file over Slack

ricktaylor:
This should close because the peer-dedup race is being addressed at a different level: the connection-pool/peer-identity plan (tcpclv4/docs/connection_pool_peer_identity.md) coalesces pools by SESS_INIT Node-ID hint as the interim step, with the real fix being BPA-level Neighbour→Peer resolution (BP-ARP) rather than a registry-lock change inside tcpclv4. The lock unification here would be superseded by that restructure. Branch will be deleted with the close.

ricktaylor (changes requested):
This PR should close

PR #569, opened 2026-06-16
refactor(bpa): use SmallRng instead of ThreadRng for non-crypto RNG

ricktaylor:
### Summary
`bpa` has two RNG sites, neither of which needs cryptographic randomness:

- `NodeIds::default()` — invents a fallback IPN node ID. Node IDs are **public identifiers** (they're advertised), so they need to
avoid collisions, not be unpredictable.
- `BundleMemStorage::save()` — mints in-memory storage keys. These are **internal handles** that need to be unique, not
unguessable.

Both used `rand::rng()` (`ThreadRng`). This PR switches them to `SmallRng`, seeded once from the OS (`SysRng`), and drops `bpa`'s
now-unused `rand/thread_rng` feature.

### Why
- **Right tool for the job:** `SmallRng` (xoshiro) is a fast, non-cryptographic PRNG — appropriate for
uniqueness/collision-avoidance. Cryptographic randomness in Hardy lives in `bpv7` BPSec, which already uses `SysRng` and is
untouched here.
- **Smaller surface:** removes the reseeding `ThreadRng` machinery — and the `chacha20` backend — from `bpa`, sidestepping the `rand` `ThreadRng` soundness advisory class entirely.

### Changes
- `bpa/src/node_ids.rs` — seed a `SmallRng` from `SysRng` for the fallback node ID.
- `bpa/src/storage/bundle_mem.rs` — hold a `SmallRng` in `Inner`, seeded once at construction; draw storage keys from it under the existing lock.
- `bpa/Cargo.toml` — drop `rand/thread_rng` (keep `sys_rng` for seeding).

### DIFF
--- a/bpa/Cargo.toml
+++ b/bpa/Cargo.toml
@@ -25,7 +25,6 @@ std = [
     "hardy-eid-patterns/std",
     "tracing/std",
     "rand/std",
-    "rand/thread_rng",
     "thiserror/std",
     "futures/std",
     "serde?/std",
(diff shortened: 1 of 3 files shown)


sylvain-pierrot:
Looks good!
Do the others bundle storage implementations need to be updated as well?

ricktaylor:
No - this is a small focussed fix to remove some 'std' dependencies in the bpa - no need to touch the serious storage impls

sylvain-pierrot (approved):
LGTM

PR #570, opened 2026-06-17
fix(bpa/rib): wait on unresolvable via instead of dropping bundles

sylvain-pierrot:


ricktaylor (changes requested):
✅ Clean — no correctness bugs. Mergeable as-is; the notes below are optional polish.

I traced the recursion change by hand and built/ran the branch (clippy -D warnings clean, 24/24 RIB tests pass incl. the new ones). The behavioural change is sound: loop detection now waits (trace! + continue) instead of dropping, with correct trail backtracking and
guaranteed termination; the Via(null) / Via(<own-node>) reject checks run before any mutation (no partial insert); is_local handles every Eid variant; all callers handle the new Result; and the gRPC error path is non-exhaustive-safe. This also lands the earlier
review's bool → Result<bool> + two-error-variant suggestion faithfully.

The remaining items are all minor.

1. Fold same-crate imports (imports_granularity = "Crate")

Dropping the use super::* globs is the right move — just collapse the per-submodule lines while you're in here. rib/mod.rs already shows the target form.

bpa/src/rib/find.rs (imports + the two use super::… lines):
use hardy_bpv7::{bundle::Bundle as Bpv7Bundle, eid::Eid, status_report::ReasonCode};
// ...
use super::{FindResult, Rib, RouteTable, route::Action};

bpa/src/rib/route.rs (top of file and the test module repeat this):
use core::{cmp::Ordering, fmt};

use hardy_bpv7::{eid::{Eid, NodeId}, status_report::ReasonCode};

bpa/src/routes.rs (the public module — higher-visibility place to keep the style consistent):
use hardy_bpv7::{eid::{Eid, NodeId}, status_report::ReasonCode};

bpa/src/rib/find.rs test module (most minor — take it or leave it):
use crate::{
    rib::route::{self, tests::{add_route, make_rib}},
    services::tests::NullService,
};

2. Consistent field visibility — bpa/src/rib/agent.rs

name is pub but agent is pub(super), though both are only read within the rib tree (mod.rs needs only the Agent type visible, as a HashMap value). Make them match:
pub struct Agent {
    pub(super) agent: Arc<dyn routes::RoutingAgent>,
    pub(super) name: String,
}
(Widening the struct itself pub(crate) → pub is fine — mod rib is private, so plain pub is correct.)

3. Document the deliberate non-tracking of rejected routes — bpa-server/src/static_routes/agent.rs

A rejected route is intentionally left out of the tracked routes Vec, so each reload re-attempts and re-warns rather than recording a route the RIB refused as "installed" (which would re-hide exactly the divergence this PR surfaces). Worth a line so it doesn't read
as an oversight:
// Rejected routes are deliberately NOT added to `accepted`/`routes`: keeping
// them out of the tracked set means each reload re-attempts and re-warns,
// instead of silently treating a route the RIB refused as installed.
let mut accepted = Vec::new();
Separately (pre-existing, not for this PR): the remove side still .ok()-swallows failures then retain-evicts unconditionally, so add/remove accounting is now slightly asymmetric — flagging only for the record.

sylvain-pierrot:
> ✅ Clean — no correctness bugs. Mergeable as-is; the notes below are optional polish.
> 
> I traced the recursion change by hand and built/ran the branch (clippy -D warnings clean, 24/24 RIB tests pass incl. the new ones). The behavioural change is sound: loop detection now waits (trace! + continue) instead of dropping, with correct trail backtracking and guaranteed termination; the Via(null) / Via() reject checks run before any mutation (no partial insert); is_local handles every Eid variant; all callers handle the new Result; and the gRPC error path is non-exhaustive-safe. This also lands the earlier review's bool → Result + two-error-variant suggestion faithfully.
> 
> The remaining items are all minor.
> 
> 1. Fold same-crate imports (imports_granularity = "Crate")
> 
> Dropping the use super::* globs is the right move — just collapse the per-submodule lines while you're in here. rib/mod.rs already shows the target form.
> 
> bpa/src/rib/find.rs (imports + the two use super::… lines): use hardy_bpv7::{bundle::Bundle as Bpv7Bundle, eid::Eid, status_report::ReasonCode}; // ... use super::{FindResult, Rib, RouteTable, route::Action};
> 
> bpa/src/rib/route.rs (top of file and the test module repeat this): use core::{cmp::Ordering, fmt};
> 
> use hardy_bpv7::{eid::{Eid, NodeId}, status_report::ReasonCode};
> 
> bpa/src/routes.rs (the public module — higher-visibility place to keep the style consistent): use hardy_bpv7::{eid::{Eid, NodeId}, status_report::ReasonCode};
> 
> bpa/src/rib/find.rs test module (most minor — take it or leave it): use crate::{ rib::route::{self, tests::{add_route, make_rib}}, services::tests::NullService, };
> 
> 2. Consistent field visibility — bpa/src/rib/agent.rs
> 
> name is pub but agent is pub(super), though both are only read within the rib tree (mod.rs needs only the Agent type visible, as a HashMap value). Make them match: pub struct Agent { pub(super) agent: Arc, pub(super) name: String, } (Widening the struct itself pub(crate) → pub is fine — mod rib is private, so plain pub is correct.)
> 
> 3. Document the deliberate non-tracking of rejected routes — bpa-server/src/static_routes/agent.rs
> 
> A rejected route is intentionally left out of the tracked routes Vec, so each reload re-attempts and re-warns rather than recording a route the RIB refused as "installed" (which would re-hide exactly the divergence this PR surfaces). Worth a line so it doesn't read as an oversight: // Rejected routes are deliberately NOT added to `accepted`/`routes`: keeping // them out of the tracked set means each reload re-attempts and re-warns, // instead of silently treating a route the RIB refused as installed. let mut accepted = Vec::new(); Separately (pre-existing, not for this PR): the remove side still .ok()-swallows failures then retain-evicts unconditionally, so add/remove accounting is now slightly asymmetric — flagging only for the record.

Thanks, I’ve addressed your points 🙂 
Let me know if you have others remarks

ricktaylor:
Thanks — imports, visibility, and the comment all look good. One small leftover from the same pass: the test-module imports in rib/route.rs and rib/find.rs are still split per-submodule, so they don't match the folding you applied at module level. Worth tidying while
we're here so the convention is consistent throughout the file.

bpa/src/rib/route.rs (test module):
```rust
use core::{num::NonZeroUsize, time::Duration};

use hardy_bpv7::{
    bundle::{Bundle as Bpv7Bundle, Id as BundleId},
    creation_timestamp::CreationTimestamp,
    eid::IpnNodeId,
};
```

bpa/src/rib/find.rs (test module):

```rust
use hardy_bpv7::{
    bundle::{Bundle as Bpv7Bundle, Id as BundleId},
    creation_timestamp::CreationTimestamp,
    eid::{IpnNodeId, NodeId, Service as EidService},
};
```

Non-blocking — purely cosmetic and not fmt-enforced. Happy to approve with or without it.

ricktaylor (approved):
LGTM!!

PR #579, opened 2026-06-22
Feat/fuzz ci

ricktaylor:
## Continuous fuzzing in CI via ClusterFuzzLite

Runs Hardy's existing cargo-fuzz targets continuously in CI using
[ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) (the OSS-Fuzz
engine hosted in GitHub Actions), chosen over a hand-rolled cargo-fuzz matrix
because reproducible crashes are a priority and CFLite is an OSS-Fuzz on-ramp.

### What's added
- **`.clusterfuzzlite/`** — `Dockerfile` (OSS-Fuzz `base-builder-rust`), `build.sh`,
`README.md` (operator runbook).
- **`.github/workflows/cflite-pr.yml`** — per-PR `code-change` fuzzing (300s, address
sanitizer), starts warm from the persisted corpus.
- **`.github/workflows/cflite-cron.yml`** — nightly Batch → Prune → Coverage.
- **Persistent corpus** in sibling repo `ricktaylor/hardy-fuzz-corpus` under the
`corpus/<target>/` layout CFLite expects; cron grows/prunes it, coverage reports
publish to its `gh-pages`.
- **No hand-maintained target list:** `build.sh` and `run_lcov.sh` both derive the
8 fuzz targets from each `*/fuzz/Cargo.toml` `[[bin]]`, so nothing can drift.

### Incidental fixes (surfaced by the fuzz build)
The fuzz build compiles `bpv7` standalone with a reduced feature set, which exposed
two latent issues that the workspace build masked via feature unification:
- `bpv7/src/editor.rs`: `std::cmp::Reverse` → `core::cmp::Reverse` (bpv7 is `no_std`).
- `bpv7/src/editor.rs`: gate `set_bib_target`/`set_bcb_target` behind `#[cfg(feature = "bpsec")]`
to match their only callers (`signer`/`encryptor`). Fuzzing covers the user-controlled
parse/verify path, not bundle creation, so `bpv7`-without-`bpsec` in the fuzzer is intended.

### Removed / updated
- Deletes `scripts/run_fuzz.sh` (CFLite owns continuous fuzzing; ad-hoc local fuzzing is
`cargo +nightly fuzz run …`).
- `run_lcov.sh` reports fuzz coverage only for targets with a real corpus (never a faked
number).
- `docs/test_strategy.md` + `docs/test_coverage_report.md`: correct the fuzz-target count
11 → 8 (uprev 1.3).

### Notes for reviewers
- The history contains a `$WORK` build-cache experiment (`65ca770c`) and its revert
(`8b9c1801`) — they cancel out, so the **net diff is unaffected**. Investigation
conclusion: CFLite's docker-in-docker sandbox wipes the nested `$WORK` after the build
and doesn't forward the GHA cache creds inward, so the ~6 min build can't be cached
here. (Squash/drop those two commits if a clean history is preferred.)

### Validation
- ✅ PR workflow is **green** end-to-end (build + corpus pull + fuzz).
- ⏳ Corpus **push-back** is exercised only by the Cron Batch/Prune jobs, which require the
workflows to be on `main` first (`workflow_dispatch` only appears there).

### Post-merge operational steps
1. Add the `FUZZ_CORPUS_ACCESS_TOKEN` repo secret (fine-grained PAT, `contents: write` on
    `hardy-fuzz-corpus`).
2. Run **ClusterFuzzLite Cron** → confirm a fresh corpus commit lands on `hardy-fuzz-corpus`.
3. Enable GitHub Pages (`gh-pages`) on `hardy-fuzz-corpus` after the first Coverage run
    creates the branch.


.github/workflows/cflite-cron.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Run Fuzzers (coverage)
+        uses: google/clusterfuzzlite/actions/run_fuzzers@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fuzz-seconds: 600
+          mode: "coverage"
+          sanitizer: coverage
+          storage-repo: ${{ env.STORAGE_REPO }}
+          storage-repo-branch: main
+          storage-repo-branch-coverage: gh-pages
```
sylvain-pierrot: I’d suggest to add “coverage” in the matrix :)

.github/workflows/cflite-cron.yml
```diff
@@ -0,0 +1,76 @@
...
+          language: rust
+
+      - name: Run Fuzzers (prune)
+        uses: google/clusterfuzzlite/actions/run_fuzzers@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fuzz-seconds: 600
+          mode: "prune"
+          storage-repo: ${{ env.STORAGE_REPO }}
+          storage-repo-branch: main
```
sylvain-pierrot: I think you could add this in the matrix as well

.github/workflows/cflite-cron.yml
ricktaylor: Done — folded batch, prune, and coverage into a single Fuzz job driven by matrix.include over {mode, sanitizer, fuzz-seconds}. One build+run step pair now covers all three. I also added max-parallel: 1: batch and prune both push the corpus main branch, so running the legs one at a time avoids a racing push to hardy-fuzz-corpus (coverage only touches gh-pages).

.github/workflows/cflite-cron.yml
ricktaylor: 👍 Prune is now a leg of that same unified matrix (alongside batch and coverage), so it's no longer a duplicated job block. Serialized via max-parallel: 1 to keep the batch/prune corpus pushes from colliding.

sylvain-pierrot: Amazing! Looks good

sylvain-pierrot (approved):
Really great improvement. Fuzz will be very useful!!
LGTM

PR #589, opened 2026-06-29
Interop: pin & version-capture all targets; converge bp ping/echo with the echo-service draft

ricktaylor:
# Summary

Two related strands of interop work, now based directly on main (the v0.2.0 release bump is no longer in this branch's history):

1. Interop suite hardening — every foreign implementation is pinned to a known-good version, build provenance/version is captured and folded into the results table, and several per-target harnesses were fixed or modernised.
2. bp ping + built-in echo-service convergence with the echo-service draft (draft-taylor-dtn-echo-service) — drop BIB signing in favour of byte-for-byte payload integrity, and make the reflector emit a freshly-sourced, payload-only response.

All eight implementations (Hardy + 7 peers) now pass at 0% loss — see tests/interop/interop_results.md.

# Interop suite hardening

- Pin every target to a known-good version and capture build provenance; the results table now carries a Version column (with declared: vs. observed where they differ) and moves the date to the header.
- NASA cFS: adopt cFS v7.0.1 (handle the cmdUtil→cmd_send rename), fix framework pins, and add a real reverse-direction TEST 2 via a new ping_app (so cFS actively pings Hardy, not just reflects).
- ESA-BP: build from current source (3.0.0) instead of the 2023 tag, with a strip-proprietary.sh helper and STCP-CLE/EchoService updates.
- DTNME: derive a declared: version from dtn-version.dat.
- dtn7-rs: pin to the XFER_ACK fix and make the harness survive its peer-reaper (previously caused ~88% spurious loss).
- Harness: per-implementation ping timeout now scales with --count; README/coverage drift corrected.

# bp ping + echo-service convergence

- echo-service: builds a fresh response bundle (new creation timestamp, payload-only reflection) instead of editing the request in place; ignores null-source bundles and administrative records; sets response control flags per the draft (response_flags) — reflect "must not be fragmented", mirror status-report-request + "status time requested" flags when reports were requested, clear app-ack, default the rest.
- bp ping: drop BIB-HMAC signing; verify round trips by retaining the sent payload and comparing it byte-for-byte against the reflection (no key management; works against any conformant echo).
- bp ping bug fix: count-mode (-c) termination released multiple semaphore permits per ping (the reply plus the deletion report Hardy emits after every local delivery, on both legs), so runs could end early and overstate loss. Each sequence number now resolves exactly one permit.

# ⚠️  User-facing change

bp ping --no-sign is removed (integrity is now automatic and keyless). Any scripts passing --no-sign must drop it. All in-repo callers (interop scripts, tvr and bpa-server ping tests, docs) are updated.

# Validation

- Full suite green: tests/interop/run_all.sh → all 8 implementations 0% loss, 20/20.
- cargo fmt --check, cargo clippy --all-targets -- -D warnings, cargo test pass for hardy-tools and hardy-echo-service.

# Notes for reviewers

- No public Rust API changes; the only interface change is the removed CLI flag.
- interop_results.md is a generated artifact (committed as the current benchmark).


bpa-server/Dockerfile
```diff
@@ -26,7 +26,7 @@ COPY . .
 RUN cargo build --release --bin hardy-bpa-server $CARGO_ARGS
 
 # ── Stage 4: minimal runtime image ───────────────────────────────────────────
-FROM gcr.io/distroless/cc-debian12 AS runtime
+FROM gcr.io/distroless/cc-debian13@sha256:a017e74bd2a12d98342dbecd33d121d2b160415ed777573dc1808969e989d94d AS runtime
```
sylvain-pierrot: Is the usage of a digest intentional here?

tcpclv4-server/Dockerfile
```diff
@@ -26,7 +26,7 @@ COPY . .
 RUN cargo build --release --bin hardy-tcpclv4-server $CARGO_ARGS
 
 # ── Stage 4: minimal runtime image ───────────────────────────────────────────
-FROM gcr.io/distroless/cc-debian12 AS runtime
+FROM gcr.io/distroless/cc-debian13@sha256:a017e74bd2a12d98342dbecd33d121d2b160415ed777573dc1808969e989d94d AS runtime
```
sylvain-pierrot: Same question here

tests/interop/docs/test_plan.md
```diff
@@ -92,6 +92,10 @@ Uses Hardy's built-in TCPCLv4 CLA. Both nodes on Docker bridge network.
 
 Uses Hardy's standalone `mtcp-cla` binary (`tests/interop/mtcp/`), which provides STCP framing (4-byte length prefix) or MTCP framing (CBOR byte string). Both nodes on same host via `--network host`.
 
+### Base images
+
+Every peer builds and runs in Docker on a single digest-pinned base (`debian:trixie-slim`), so OS and glibc are constant across the comparison; DTNME (Debian buster) and ud3tn (`python:3.13-slim`) are the documented exceptions. The Hardy baseline runs its echo from the shipping distroless `bpa-server` image. See [README § Docker images](../README.md#docker-images).
```
sylvain-pierrot, line 97: Great!

tests/interop/ud3tn/docker/Dockerfile
```diff
@@ -14,9 +14,9 @@
 #       ud3tn-interop \
 #       -e ipn:2.0 -c "mtcp:0.0.0.0,4557" -b 7 -A 0.0.0.0 -P 4243 -R
 
-FROM python:3.13-slim AS builder
+FROM python:3.13-slim@sha256:2b7445fb71ca9cb15e9aab053fe8cb3162796f8e1d92ada12a49c766a811bc1e AS builder
```
sylvain-pierrot: Digest??

sylvain-pierrot (commented):
Great job. I will have some others comments.

tests/interop/ud3tn/docker/Dockerfile
ricktaylor: Yeah - doesn't need to be on the build image

bpa-server/Dockerfile
ricktaylor: Yes it is - to make sure we have a completely reproduceable build

tests/interop/ud3tn/docker/Dockerfile
ricktaylor: Fixed

bpa-server/Dockerfile
sylvain-pierrot: OK, thanks. I understand the reproducibility concern, but I have two questions:
- Without pinning the digest, what exactly makes the build non-reproducible?
- Since `bpa-server` is intended for production use (not just testing), wouldn't pinning the digest also prevent us from automatically picking up security fixes (e.g. CVE patches, or whatever updates/fixes) published in newer distroless images?

ricktaylor: Ok - so production is no longer pinned, but the tests are

sylvain-pierrot (approved):
LGTM! Great improvement for testing

PR #591, opened 2026-07-03
Bump MSRV to 1.95 and enforce it in CI

ricktaylor:
# Summary

Raises the workspace MSRV from 1.87 → 1.95, adds a CI job that actually enforces it, and applies the cleanups the newer baseline enables (let-chains, the built-in cfg_select! macro, and the resulting clippy/fmt tidy).

# Motivation

rust-version in Cargo.toml declares a support baseline, but nothing enforced it — the build/test jobs run on latest stable, so code using a feature newer than the declared MSRV would compile in CI yet break for anyone actually on that MSRV. This PR makes the baseline
real: it moves it up to where the code wants to be (let-chains, cfg_select!) and adds a job that fails if the workspace stops building on the declared version.

# What's in it

- MSRV bump — rust-version = "1.95" (via 1.94 → 1.95).
- CI enforcement — new msrv job reads rust-version from Cargo.toml, installs that toolchain, and runs cargo check --locked --all-features --workspace.
- Let-chain adoption — collapse nested if let/if let Err into if let … && let … where clippy's collapsible_if now expects it (bpa filter/registry, storage backends, tcpclv4).
- cfg_select! — replace the cfg-if crate with the built-in macro in the bpa-fuzz harness's storage-backend selection, and drop the cfg-if dependency.
- cargo fmt across the touched crates.

# Files

 .github/workflows/rust.yml                                         │ +msrv validation job
Cargo.toml / Cargo.lock                                            │ MSRV 1.87→1.95; drop cfg-if
bpa/src/filter/{rfc9171,validity}.rs, bpa/src/services/registry.rs │ let-chain collapse
localdisk-storage, s3-storage, tcpclv4/src/connection.rs           │ let-chain collapse
bpa/fuzz/{Cargo.toml,src/lib.rs}                                   │ cfg_if! → cfg_select!
bpa/benches/bundle_bench.rs, bpa/tests/pipeline.rs                 │ fmt                         

# Notes

- Split out of the interop PR (feat/interop-version-capture), which is intentionally kept at 1.87. Merge this first (or rebase the interop branch on it): once CI's clippy is new enough to flag collapsible_if, the interop PR needs these fixes to stay green.
- No behavioural changes — purely toolchain baseline, lint style, and a dependency swap. Validated by the standard fmt/clippy/build/test gates plus the new msrv check.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -139,3 +139,28 @@ jobs:
         run: |
           ./target/debug/hardy-bpa-server &
           bash ./tests/interop/hardy/run_hardy_ping_test.sh --peer 127.0.0.1:4556 --count 5
+
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Read MSRV from Cargo.toml
+        id: msrv
+        run: echo "version=$(grep -m1 '^rust-version' Cargo.toml | sed -E 's/.*"([^"]+)".*/\1/')" >> "$GITHUB_OUTPUT"
+
+      - name: Setup Rust toolchain (MSRV ${{ steps.msrv.outputs.version }})
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: ${{ steps.msrv.outputs.version }}
+
+      - name: Use cached dependencies and artifacts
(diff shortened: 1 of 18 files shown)


sylvain-pierrot (dismissed):
Very useful in CI. LGTM!

sylvain-pierrot (approved):

PR #592, opened 2026-07-03
Fix tcpclv4 fuzz coverage runs; read fuzz coverage from GitHub Pages

ricktaylor:
## Summary
Two related coverage changes:
1. **Fix the tcpclv4 fuzz harness** so the ClusterFuzzLite coverage runs actually complete.
2. **Make `run_lcov.sh` read fuzz coverage from the published GitHub Pages dashboard** instead of recomputing it locally.

(2) is what surfaced (1): switching the summary to read the published per-target
stats exposed that tcpclv4's `active` target had no published coverage and
`passive` had collapsed to ~5%.

## `fix(tcpclv4-fuzz)` — coverage runs complete again
The `active` target accepted only one connection per iteration, but the CLA's
`connect()` retries several times. Retries 2..n hit an unanswered socket and each
waited out the full contact timeout, so one iteration took ~8s+. Replaying the
whole corpus under the coverage job's time budget was interrupted — `active`
published no coverage at all, and `passive` was slow for the same timeout reason.

- `active`: accept and serve **every** connection the connector opens (a `select!`
  loop against the connect task) — no attempt stalls, and every retry is now
  exercised with the fuzz data.
- `active`: bind an ephemeral port instead of the shared `:4556`; the target owns
  both ends, so it can't collide with `passive`.
- `lib`: drop the fuzz `contact_timeout` 2s → 1s (u16 seconds, the floor).
- Shrink per-iteration backstops (active 15s → 5s, passive 5s → 3s) now that
  iterations finish in milliseconds.

Verified: both targets now replay the full corpus at real `exec/s` and produce
complete coverage (`active` completes instead of being interrupted; `passive`
recovers from the degenerate 5.3%).

## `build(coverage)` — read fuzz coverage from GitHub Pages
`run_lcov.sh` now reads each target's published llvm-cov export
(`coverage/latest/fuzzer_stats/<target>.json`) from the CFLite GitHub Pages site
and filters it to the crate's own `src/`, instead of recomputing with
`cargo +nightly fuzz coverage`. Faster, needs no nightly toolchain or corpus
checkout, and matches the dashboard exactly.

- Unit coverage is still measured locally (`cargo llvm-cov`).
- Fuzz source overridable via `HARDY_COVERAGE_URL`.
- A target present in-tree but missing from the published run is reported as
  **UNAVAILABLE** (with a warning) — never silently dropped. (This is the guard
  that caught the `active` gap.)

## Testing
- `cargo check` and `cargo clippy --all-targets -- -D warnings` on the fuzz crate: clean.
- `run_lcov.sh` exercised end-to-end (happy path + the UNAVAILABLE guard).
- Manual `cargo +nightly fuzz coverage active/passive`: both complete.

## Notes
- `docs/coverage_summary.md` is intentionally **not** included; regenerate it from
  a full run (with `cargo-llvm-cov`) once CI republishes the fixed tcpclv4 coverage.
- Stale `fuzz/artifacts/{active,passive}/crash-*` reproducers predate this and
  should be triaged separately.


sylvain-pierrot (approved):
Approved!

PR #593, opened 2026-07-05
Enforce timeout for bp-ping

hugoponthieu:
## Summary

This PR makes `bp ping --timeout` enforce a real wall-clock session deadline instead of only influencing bundle lifetime calculation.

The ping runner now races the active ping session against the configured timeout, cancels the session when the deadline expires, and still prints the normal ping summary from the shared service statistics.

The interop scripts that invoke Hardy's `bp ping` now pass the native `--timeout` option. Existing shell `timeout` wrappers around Hardy's own `bp ping` invocations were removed where present, while timeouts around third-party peer tools were left unchanged.

While checking those interop scripts, this PR also fixes two test-environment issues that prevented the updated coverage from running reliably:

- the ION Docker builder now installs the Autotools macro dependencies needed by current ION sources (`pkg-config` and `autoconf-archive`);
- the NASA cFS startup script uses the installed cFS command sender, no longer hides BPNode command failures, and its timeout path prints STCP/contact diagnostics when port `4501` never opens.

## Behavior

Before this change, `bp ping -w/--timeout` was documented as a total session time limit but did not stop the ping session.

After this change, `--timeout T` caps the total `bp ping` runtime for both finite and count-less sessions. On timeout, pinging stops, the partial summary is printed, and the command exits according to the responses received so far.

## Testing

Validated locally:

```sh
cargo test -p hardy-tools
cargo clippy -p hardy-tools --all-targets --all-features -- -D warnings
```

Interop commands to run where the required peer implementation is available:

```sh
./tests/interop/hardy/test_hardy_ping.sh
./tests/interop/hardy/run_hardy_ping_test.sh
./tests/interop/ION/test_ion_ping.sh
./tests/interop/DTNME/test_dtnme_ping.sh
./tests/interop/ESA-BP/test_esa_bp_ping.sh
./tests/interop/HDTN/test_hdtn_ping.sh
./tests/interop/ud3tn/test_ud3tn_ping.sh
./tests/interop/dtn7-rs/test_dtn7rs_ping.sh
./tests/interop/NASA-cFS/test_cfs_ping.sh
```

Manual smoke test against a local Hardy BPA with echo service enabled:

```sh
cargo run -p hardy-bpa-server -- -c bpa-server/examples/config-tcp-1.yaml
cargo run -p hardy-tools -- ping -c 5 --timeout 20s --no-sign ipn:1.7 127.0.0.1:4556
```

Observed result:

```text
5 bundles transmitted, 5 received, 0% loss
```

## Notes

This PR intentionally does not change shell timeouts around peer implementation commands such as `bping`, `ping_me`, or `aap2_ping`, because Hardy's `bp ping --timeout` does not affect those tools.


ricktaylor:
Hi Hugo,

Thank you so much for this.  I have recently refreshed some of the interop test scripts and procedures, so while I review what you have, could you rebase onto main please?

Rick

hugoponthieu:
Hi Rick,

I rebased and it should be good to go. Tell me if you see anything else ! 

Hugo

ricktaylor:
Hi Hugo,

Thanks for the rebase - one quick thing I noticed: `--no-sign` is no longer an option for `bp ping` (BPSec doesn't actually help, so the option has been removed).

Can you fix, and I'll give you a full review after?

hugoponthieu:
Hi Rick, 

Did removed occurences of --no-sign and rebased. Tell me if you see anything else before full review.

ricktaylor:
Hi Hugo - can you sign your commits please?

ricktaylor (changes requested):
Hi Hugo,

Thanks for coming back to this after the merge of `main` — the branch is now level with `main` (merge-base is the current tip), `cargo clippy -p hardy-tools --all-targets --all-features -- -D warnings` and `cargo fmt --check` are both clean, and the restructuring so that the shared `service::Service` outlives the session means the partial summary now prints correctly on timeout. That was the trickiest piece of the previous feedback and it's done right. `--timeout` is now genuinely enforced for both finite and count-less sessions, which is what this PR set out to do.

There are two regressions that came in with the merge, one design point on the timeout mechanism, and two behavioural risks I'd like addressed before this lands. Details below.

## Blocking

### 1. cFS TEST 1 readiness check regressed against `main`

`main` (5eabd812) introduced the shared `wait_for_port` helper in `tests/interop/lib/wait.sh`, and cFS TEST 1 in `tests/interop/NASA-cFS/test_cfs_ping.sh` used it. The merge kept this branch's older inline `ss -tln` loop instead, and it has three problems:

- **A dead container falls through into the test.** When `docker ps` shows the container gone, the loop `break`s out of the `while` with `WAIT_COUNT` still below `WAIT_TIMEOUT`, so the `if [ $WAIT_COUNT -ge $WAIT_TIMEOUT ]` guard below it does not fire. The script then logs "cFS is accepting connections" and runs TEST 1 against a container that has already exited. `wait_for_port` returns 2 in this case and the test is skipped with the logs dumped.
- **The `ss` rationale is the one `wait_for_port` was written to replace.** The comment says "nc -z would be accepted by the CLA", but the helper's doc comment covers this: the probe opens and immediately closes the socket without writing, so it feeds no stray bytes to a length-prefixed STCP peer — and unlike `ss -tln | grep`, a real connect can't false-positive on a bound-but-not-yet-accepting socket and doesn't depend on Docker's port-publishing mode.
- **TEST 2 in the same script still uses `wait_for_port`**, so the two directions now use different readiness mechanisms.

Please restore the `wait_for_port` call for TEST 1. The one thing in the new block genuinely worth keeping is the improved failure diagnostics — the filtered `docker logs | grep -i 'BPNODE\|STCP\|...'` dump — which fits nicely in the failure branch of the `wait_for_port` version.

### 2. `tests/interop/mtcp/Cargo.lock` — unrelated churn, please revert

This PR changes no mtcp sources, but the lockfile diff (105+/38−) *downgrades* `pest` 2.8.7→2.8.6, `wasm-bindgen` 0.2.126→0.2.122, `js-sys`, `zerocopy`, `rapidhash`, and `wasip2` — and the `pest` downgrade drags in a duplicate old-generation crypto stack (`sha2` 0.10, `digest` 0.10, `block-buffer` 0.10, `generic-array` 0.14, `crypto-common` 0.1). This looks like a regeneration against a stale local registry. Please restore the file to `main`'s version (`git checkout origin/main -- tests/interop/mtcp/Cargo.lock`).

## The timeout mechanism itself

The current shape in `run_ping` (`tools/src/ping/exec.rs`):

```rust
match args.timeout {
    Some(t) => tokio::select! {
        r = exec_res => Some(r?),
        _ = tokio::time::sleep(*t) => { tasks.cancel_token().cancel();
            None
             }
    },
    None => Some(exec_res.await?),
};
```

Three separate things worth untangling here.

**First, `tokio::time::timeout()` is the cleaner wrapper for a plain two-way race.** The hand-rolled `select!` + `sleep` is exactly what `timeout()` encapsulates, and the wrapper avoids the incidental problems the macro form has picked up: the `match` currently produces a discarded `Some`/`None` value that exists only to satisfy the arms, and the timeout arm's formatting is hand-rolled — which rustfmt will never fix, because it doesn't reformat `select!` interiors, so it slips past `cargo fmt --check` while still being the kind of hand-formatting the code style guide (`docs/style_guides/code_style_guide.md`, § Formatting) rules out. As a straight substitution:

```rust
match args.timeout {
    Some(t) => {
        if let Ok(r) = tokio::time::timeout(*t, exec_res).await {
            r?;
        }
    }
    None => exec_res.await?,
}
```

(For contrast, the three-way race in `exec_external_cla` — deadline vs child exit vs startup delay — is a legitimate `select!` and should stay as it is.)

**Second, however, both forms share the same flaw: they drop the session future.** On expiry, `exec_inner` is dropped wherever it happens to be suspended. Dropped mid-`sink.send()`, a bundle can be on the wire but never recorded, so the summary undercounts `transmitted`. Nothing inside `exec_inner` gets a chance to observe cancellation — and note that in the current code the `.cancel()` in the timeout arm fires only after the race is already decided, so it cannot help `exec_inner`; it only reaches the signal listener task.

**Third, the machinery for doing this cleanly already exists in this function — and it's the project convention.** The send loop breaks on `cancel_token.cancelled()` between sends, and `wait_for_responses` selects on the token — that's how Ctrl+C already stops the session, via `listen_for_cancel`. The code style guide (`docs/style_guides/code_style_guide.md`, § Async and Concurrency) makes cancellation tokens the standard shutdown mechanism, with `hardy-async` primitives preferred over reaching for `tokio` directly. A deadline is just a scheduled Ctrl+C, so the timeout should ride the same path:

```rust
if let Some(t) = args.timeout {
    let cancel = tasks.cancel_token().clone();
    hardy_async::spawn!(tasks, "ping_deadline", async move {
        if tokio::time::timeout(*t, cancel.cancelled()).await.is_err() {
            cancel.cancel();
        }
    });
}

exec_inner(args, bpa.as_ref(), tasks.cancel_token(), service.clone()).await?;
```

With this, `exec_inner` unwinds cooperatively: the in-flight send completes and is counted, the wait loop exits, it returns `Ok`, and the summary/shutdown sequence is identical for Ctrl+C, deadline, and normal completion — one code path instead of three. Wrapping the sleep in `timeout(*t, cancel.cancelled())` also means the deadline task itself exits promptly when the pool shuts down after a normal run, rather than holding a bare `sleep`.

The one thing the drop-the-future approach buys is a hard bound even if `sink.send()` itself wedges — but in that state the subsequent `bpa.shutdown().await` would hang anyway, so it doesn't actually deliver a bounded process exit (see item 4).

## Behavioural risks

### 3. Internal timeout can convert a partial run into an interop PASS

The interop scripts pass on `RECEIVED == TRANSMITTED`; they never compare against `PING_COUNT`. The old external `timeout` killed the process with exit 124 → FAIL. The new internal timeout exits 0, so a session that only got 3 of 10 pings away before the deadline — with all 3 answered — prints `3 bundles transmitted, 3 received, 0% loss` and passes.

The tidiest fix is tool-side, and it matches the "exit codes matching Linux/BSD ping conventions" comment already in exec.rs: iputils `ping -c N -w D` exits 1 when the deadline arrives with fewer than N replies received. If `run_ping` does the same (count set, timeout expired, `received < count` → `NoResponse`), every script fails correctly with no script changes at all.

### 4. Removing the shell `timeout` leaves startup and shutdown unbounded

`--timeout` bounds the ping session, but not what surrounds it: BPA construction, CLA startup, and — after the deadline fires — `tasks.shutdown().await`, `bpa.shutdown().await`, and the external-CLA child cleanup. A wedged peer is precisely the scenario these interop timeouts exist for, and with the outer `timeout $((...))s` removed a hung shutdown now hangs CI forever. Two cheap options, either is fine: keep the shell `timeout` as an outer belt-and-braces bound with a larger budget than `--timeout` (say `+30s`), or `std::process::exit` after the summary is printed on the deadline path.

## Non-blocking

- `print_summary()` and `statistics()` run while the BPA is still live, so a response landing between them can make the printed summary and the exit code disagree. Snapshot the statistics once and print from the snapshot, or print after `bpa.shutdown()`.
- The new `use std::sync::Arc;` sits two lines above a fully-qualified `std::sync::Arc::new(...)`; the rest of the file is consistently fully-qualified — either is fine, but pick one.
- `tests/interop/NASA-cFS/docker/start_cfs` replaces `main`'s `command -v cmd_send || command -v cmdUtil` lookup with hardcoded `/opt/cfs/host` paths; the Dockerfile already puts that directory on `PATH`, so this is lateral churn — please keep `main`'s lookup. The `send_bpnode_cmd` wrapper that warns loudly instead of `|| true` is a genuine improvement though; please do keep that.
- The `$((PING_COUNT * 2 + 10))s` budget is now repeated in eight scripts and silently assumes `--interval 1s`. A follow-up could derive a default session deadline inside the tool, mirroring `Command::lifetime()`.

## Good as-is

- `tests/interop/ION/docker/Dockerfile`: `pkg-config` + `autoconf-archive` are exactly what current ION autotools sources need.
- The `+x` mode bit on `tests/interop/hardy/run_hardy_ping_test.sh`.
- Native `--timeout` in all eight interop ping invocations, with third-party peer tool timeouts correctly left alone.

tools/src/ping/exec.rs
```diff
@@ -265,13 +265,45 @@ async fn run_ping(
     let tasks = hardy_async::TaskPool::new();
     hardy_async::signal::listen_for_cancel(&tasks);
 
-    let stats = exec_inner(args, bpa.as_ref(), tasks.cancel_token()).await?;
+    let service = std::sync::Arc::new(service::Service::new(args));
```
ricktaylor: Can we use alloc:: not std:: please?

ricktaylor (changes requested):
Looks really good - can we use alloc:: instead of std:: wherever we can - so eventually we might be able to support a no_std ping

ricktaylor:
### Minor, non-blocking (fold into any future push; none justifies a round-trip)

- The new atomics are spelled `std::sync::atomic::{AtomicBool, Ordering}`; in the same no_std-aspiration spirit as the `alloc::` request, these live in `core::sync::atomic`. (The pre-existing `std::time::Duration` sites in the external-CLA block and `std::process::exit` are inherently std-adjacent and fine as they are.)
- Cosmetic: the summary counters come from the snapshot while the lost-bundle list reads live state under the lock, so a reply landing in that window can print `received N−1` yet omit the seqno from the lost list. Harmless self-consistency wobble in a torn-down session; noted only so nobody chases it as a bug later.

ricktaylor (approved):
LTGM!!  Great stuff

ricktaylor:
Sorry - can you rebase one more time - it will fix the clippy warning (unrelated to your changes)

ricktaylor:
Hi Hugo - can you rebase with `git rebase -S main` so it get's signed

ricktaylor:
Aaah - thank you - but you need to attach your PGP key with GitHub

hugoponthieu:
Sorry, I signed from the wrong machine ! Tell me if anything else is needed.

ricktaylor:
Great job, thank you!

PR #601, opened 2026-07-08
fix: rpc-proxy reader deadlock and silent bundle loss on delivery failure

sylvain-pierrot:
## Summary

Fixes two independent bugs that together cause delivery stalls under moderate concurrent load. Reproduced with as few as 3 concurrent transfers between two `hardy-bpa`-backed peers on loopback; either bug alone is sufficient to lose messages, both together are needed to see the stall symptom.

## 1. `RpcProxy` reader deadlock (`proto`)

`reader_task` used to spawn each request handler with `pool.spawn(...).await`. When the handler pool saturated, the reader parked on that await *while holding an already-read message*, so it could no longer dequeue response messages from the same stream. Those responses were exactly what would have completed the in-flight `RpcProxy::call` waiters holding the busy permits. Deadlock.

**Fix**: reader now acquires a semaphore permit *before* `stream.message()`. Responses drop the permit (no handler needed); requests move the permit into the spawned handler via RAII, releasing on completion. Bounding semantics and backpressure preserved.

## 2. Silent bundle loss on delivery failure (`bpa`)

`Service::on_receive` and `Application::on_receive` returned `()`, so `LowLevelService::on_receive` on the proto server side collapsed every failure into a `warn!("Service refused notification")` and returned quietly. The dispatcher treated that as successful delivery and called `delete_bundle`. The bundle is silently gone; whatever depended on it has no way to recover.

Failure modes that hit this path: transport hiccup, HTTP/2 GOAWAY, remote decode error, closed pending map after the reader exits, non-Receive response envelope.

**Fix** (breaking trait change): `on_receive` now returns `services::Result<()>`. On `Err`, the dispatcher transitions the bundle to `WaitingForService { service: destination }` and calls `watch_bundle` — same pattern as `dispatcher::admin` uses for status reports. `poll_service_waiting` re-delivers on the service's next registration.

Every in-tree impl is updated in the same commit. Downstream `Service`/`Application` impls need a one-line change (`Ok(())` at the end of `on_receive`).

## Testing

- **`bpa/tests/pipeline.rs::delivery_error_keeps_bundle_in_metadata`** — sends a bundle to a `FailingApp` whose `on_receive` returns `Err`, waits for the attempt, then asserts `ApplicationSink::cancel(bundle_id)` returns `true` (bundle survived). Without the fix, `delete_bundle` fires and cancel returns `false`.
- Acceptance: 6 concurrent 400 MiB transfers between two peers using this branch, pre-fix reliably stalled at 66-88 % completion with `warn!("Service refused notification")` in server logs; post-fix all deliveries complete byte-perfect with no wedged state.

## Breaking change

`Service::on_receive` and `Application::on_receive` signatures:

```rust
// before
async fn on_receive(&self, data: Bytes, expiry: OffsetDateTime);

// after
async fn on_receive(&self, data: Bytes, expiry: OffsetDateTime)
    -> hardy_bpa::services::Result<()>;
```

Downstream impls: append `Ok(())` at the end and update the return type. Returning `Err` defers the bundle to `WaitingForService` instead of deleting it.


bpa/src/storage/channel.rs
```diff
@@ -376,14 +376,15 @@ impl Store {
...
+                    // against pool shutdown; otherwise store.shutdown() (which
+                    // awaits this task) would hang. Bundles not yet enqueued on
+                    // cancel keep their persisted status and are recovered on
+                    // next process start.
                     let send = shared_cloned.tx.send(bundle);
                     pin_mut!(send);
                     select_biased! {
                         r = send.fuse() => r.map_err(|_| ())?,
-                        _ = cancel.cancelled().fuse() => return Ok(pushed_one),
+                        _ = cancel.cancelled().fuse() => return Err(()),
```
ricktaylor: Ok(false) seems a better shape for "channel closing" than Error on a graceful shutdown.

If you need to differentiate, perform the check at the call-site

ricktaylor (changes requested):

bpa/src/services/mod.rs
```diff
@@ -117,13 +117,15 @@ pub trait Application: Send + Sync {
...
+    /// Returning `Err` parks the bundle as `WaitingForService`; a
+    /// subsequent registration on the same EID re-delivers it.
     async fn on_receive(
         &self,
         source: Eid,
         expiry: time::OffsetDateTime,
         ack_requested: bool,
         payload: Bytes,
-    );
+    ) -> Result<()>;
```
ricktaylor, line 128: Do we need a specific "Receive Failed" error, or do you conside `Internal` to be sufficient?

sylvain-pierrot: Whatever error means the delivery is not a success. I’m not sure we need a specific "Receive Failed" error. What’s your take on this?

bpa/src/storage/channel.rs
sylvain-pierrot: I’ve reverted to Ok(pushed_one)

bpa/src/services/mod.rs
ricktaylor: I think we're good to re-use Internal - it's what we do for the Cla trait.

echo-service/src/lib.rs
```diff
@@ -143,8 +143,13 @@ impl hardy_bpa::services::Service for EchoService {
     }
 
     /// Called when a bundle arrives
-    async fn on_receive(&self, data: hardy_bpa::Bytes, _expiry: time::OffsetDateTime) {
+    async fn on_receive(
+        &self,
+        data: hardy_bpa::Bytes,
+        _expiry: time::OffsetDateTime,
+    ) -> hardy_bpa::services::Result<()> {
         _ = self.echo(data).await;
```
ricktaylor: Do we need `_ = ...` and `Ok(())`  - could we not just make this one line with a `map_err()` or whatever

ricktaylor (changes requested):

proto/src/proxy.rs
```diff
@@ -241,7 +235,12 @@ where
         handler: Box<dyn ProxyHandler<SMsg = S::Msg, RMsg = R::Msg>>,
     ) -> Self {
         let tasks = hardy_async::TaskPool::new();
-        let handler_tasks = hardy_async::BoundedTaskPool::default();
+        let handler_tasks = hardy_async::TaskPool::new();
+        // Held only by the reader task's clone plus any live handler's
+        // OwnedSemaphorePermit; drops when the last reference is gone.
+        let handler_permits = Arc::new(tokio::sync::Semaphore::new(
```
ricktaylor: This is a BoundedTaskPool surely?

echo-service/src/lib.rs
sylvain-pierrot: Good catch! Fixed

bpa/tests/pipeline.rs
```diff
@@ -87,8 +87,9 @@ impl services::Application for TestApp {
         _expiry: time::OffsetDateTime,
         _ack_requested: bool,
         payload: Bytes,
-    ) {
+    ) -> services::Result<()> {
         let _ = self.received_tx.send((source, payload));
+        Ok(())
```
ricktaylor: Can we use .map_err(Into::into) here?

ricktaylor (changes requested):

bpa/tests/pipeline.rs
sylvain-pierrot: fixed

proto/Cargo.toml
```diff
@@ -17,7 +17,7 @@ crate-type = ["rlib"]
 
 [features]
 default = []
-serde = ["dep:serde"]
+serde = ["dep:serde", "serde/derive"]
```
ricktaylor: Can we add the `derive` to the features of the serde dependency. its more consistent with the rest of the codebase

ricktaylor (changes requested):

proto/Cargo.toml
sylvain-pierrot: Done :)

ricktaylor (approved):
LGTM!

PR #602, opened 2026-07-08
fix(bpa/rib): store poll_waiting wakeups instead of dropping them

ricktaylor:
notify_updated() used notify_waiters(), which stores no permit: a route event arriving while the single poll_waiting_task consumer was mid-scan was silently lost, stranding the Waiting backlog until the next route event — or lifetime expiry in a static topology. notify_one() stores the permit so the scan re-runs, matching every other single-consumer wakeup in the crate.

### DIFF
--- a/bpa/src/routing/rib.rs
+++ b/bpa/src/routing/rib.rs
@@ -459,7 +459,10 @@ impl Rib {
     }
 
     async fn notify_updated(&self) {
-        self.poll_waiting_notify.notify_waiters();
+        // notify_one() stores a permit if the poll_waiting_task is mid-scan,
+        // so a route change during an in-flight scan triggers a re-scan rather
+        // than being lost (notify_waiters() stores nothing).
+        self.poll_waiting_notify.notify_one();
     }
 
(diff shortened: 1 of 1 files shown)


bpa/src/routing/rib.rs
```diff
@@ -459,7 +459,10 @@ impl Rib {
     }
 
     async fn notify_updated(&self) {
-        self.poll_waiting_notify.notify_waiters();
+        // notify_one() stores a permit if the poll_waiting_task is mid-scan,
+        // so a route change during an in-flight scan triggers a re-scan rather
+        // than being lost (notify_waiters() stores nothing).
+        self.poll_waiting_notify.notify_one();
     }
```
sylvain-pierrot: Very nice! 
I would take avantage of this PR to inline `self.poll_waiting_notify.notify_one();` directly and get rid of this method wrapper for more clarity :) 
(exactly like line 121 in `start()`)

ricktaylor: Yeah - but it's in the hardy_async crate in order remove the [std] direct dependency - so I don't really want to change it today

sylvain-pierrot (approved):

PR #604, opened 2026-07-08
fix(bpa/rib): wake the Waiting poller on every Route addition

ricktaylor:
Rib::add only notified the poll_waiting task when adding a route preempted a non-empty peer queue. Adding a Via route toward a connected but idle peer — the normal TVR contact-window case — flipped no queues, so bundles already parked in Waiting for the newly-routable destination were never re-dispatched and the contact was missed.

A new route is a new forwarding opportunity regardless of queue state: notify unconditionally, as the Local/Forward arms already do. Queue preemption for impacted vias is retained as a parallel effect.

### DIFF
--- a/bpa/src/routing/rib.rs
+++ b/bpa/src/routing/rib.rs
@@ -232,26 +232,20 @@ impl Rib {
             vias
         };
 
-        let changed = match action {
-            Action::Internal(InternalAction::AdminEndpoint) => false,
-            Action::Internal(InternalAction::Local(_))
-            | Action::Internal(InternalAction::Forward(_)) => true,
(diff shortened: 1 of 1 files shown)


bpa/src/routing/rib.rs
sylvain-pierrot: We could remove the variable `changed` and move `self.notify_updated().await;` in the match

ricktaylor: Yeah - tidier

ricktaylor: Done

sylvain-pierrot (approved):
LGTM!

PR #605, opened 2026-07-08
fix(bpa/dispatcher): return bundle to Waiting when CLA forward fails

ricktaylor:
The Forward-arm error path in process_bundle only re-armed the reaper expiry watch; it never reset the persisted status. A bundle caught by a peer disconnect between the RIB lookup and the forward stayed in Dispatching (or ForwardPending for a retired peer) — states that poll_waiting never scans — so it sat invisible until lifetime expiry instead of being re-routed on the next route event.

Set the status to Waiting before watching, as the no-route arm and the forward_bundle failure paths already do.


sylvain-pierrot (approved):

sylvain-pierrot:
LGMT!

PR #606, opened 2026-07-08
fix(bpa/storage): validate fragment coverage and 32-bit range safety …

ricktaylor:
…in reassembly

reassemble() validated fragment completeness by summing payload lengths against total_adu_length, so overlapping fragments whose lengths happened to sum to the total passed the check and a silently corrupt ADU (zero-filled gaps, double-written overlaps) was delivered to the local service. Replace the sum with a tiling check: sorted fragment extents must cover [0, total_adu_length) contiguously with no overlap.

total_adu_length and fragment offsets were also narrowed with `as usize`, truncating wire-derived u64 values on 32-bit targets and letting bounds checks pass on wrapped values. All range arithmetic is now done in u64, narrowing via try_from only after the value is proven in range.

The strict tiling check remains non-conformant with RFC 9171 §5.9 material-extents reassembly (legally overlapping fragment sets are rejected rather than trimmed), as the previous code always was outside the corrupting coincidence. That, and the absence of deletion status reports on reassembly failure, are recorded in bpa/docs/TODO.md.

### DIFF
--- a/bpa/docs/TODO.md
+++ b/bpa/docs/TODO.md
@@ -1,0 +1,25 @@
+# bpa TODO
+
+## RFC 9171 §5.9 material-extents reassembly (overlapping fragments)
+
+### Background
+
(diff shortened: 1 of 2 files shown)


sylvain-pierrot (approved):
LGTM!

PR #607, opened 2026-07-08
fix(bpa/storage): make mem-backend load() non-destructive

ricktaylor:
BundleMemStorage::load destructively popped the cache entry, on the premise that load is always the final access before delete. The forwarding retry paths break that premise: forward_bundle re-loads on every attempt, so a single transient CLA failure (session drop mid-send) destroyed the payload and the bundle was later dropped as DepletedStorage despite being unexpired — permanent data loss from a routine link flap, in the default (no backend configured) build.

load() now clones the Bytes handle (a refcount bump, not a data copy) and the entry lives until delete(). The cost is that editors rewriting loaded data fall back from in-place mutation to a copying flatten, since the retained copy must survive the rewrite — a bundle-sized memcpy per forward, in exchange for not losing the bundle.

The non-destructive contract is now documented on BundleStorage::load and enforced across all backends by BLOB-05 in the compliance suite.

### DIFF
--- a/bpa/src/storage/backend.rs
+++ b/bpa/src/storage/backend.rs
@@ -162,6 +162,11 @@ pub trait BundleStorage: Send + Sync {
     async fn recover(&self, stream: &dyn Sender<RecoveryResponse>) -> Result<()>;
 
     /// Loads the bundle stored under `storage_name`, or `None` if absent.
+    ///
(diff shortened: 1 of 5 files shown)


sylvain-pierrot (approved):
LGTM

PR #608, opened 2026-07-08
fix(bpv7): clamp creation-timestamp sequence nudge to sub-millisecond

ricktaylor:
as_datetime() cast the wire-derived u64 sequence number straight to i64 nanoseconds: any sequence >= 2^63 wrapped negative and shifted the computed creation time up to ~292 years into the past, and large positive values shifted it up to ~292 years into the future. Either way a sender-chosen sequence number corrupted downstream expiry evaluation (dropping live bundles or retaining expired ones).

The sequence nudge exists for sub-millisecond ordering — now() derives it from the nanosecond remainder within the DtnTime millisecond — so clamp it below 1ms: the result can never leave the creation millisecond, bounding the worst-case error for arbitrary wire values.

### DIFF
--- a/bpv7/src/creation_timestamp.rs
+++ b/bpv7/src/creation_timestamp.rs
@@ -98,17 +98,21 @@ impl CreationTimestamp {
 
     /// Converts the `CreationTimestamp` to a `time::OffsetDateTime`, if possible.
     ///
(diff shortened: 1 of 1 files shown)


sylvain-pierrot (approved):
Approved

PR #609, opened 2026-07-08
fix(bpa/dispatcher): keep Bundle extents consistent through forward-t…

ricktaylor:
…ime rewrite

update_extension_blocks rewrote the wire data (Previous Node insertion, Hop Count increment, Bundle Age update) but discarded the updated Bundle, feeding the Egress filter chain a Bundle whose block extents index the pre-rewrite layout. Any Egress WriteFilter using those extents (e.g. ipn-legacy-filter, a default bpa-server feature) then flattened stale ranges over the rewritten bytes, emitting a corrupt bundle on the wire — and since the CLA send succeeded, the original was deleted locally.

Use Editor::rebuild_bundle(), which returns a Bundle with extents updated to the new layout (as the parser already does), and swap it in before the Egress chain. Added a pipeline test asserting every block extent matches a re-parse of the data the filter received.

### DIFF
--- a/bpa/src/dispatcher/forward.rs
+++ b/bpa/src/dispatcher/forward.rs
@@ -15,8 +15,9 @@ impl Dispatcher {
             return;
         };
 
-        // Increment Hop Count, etc...
-        // We ignore the fact that a new bundle has been created, as it makes no difference below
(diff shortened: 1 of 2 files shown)


sylvain-pierrot (approved):
LGTM

PR #610, opened 2026-07-08
fix(bpa/filter): run filter tasks on a dedicated engine pool

ricktaylor:
run_readers spawned per-filter tasks on the caller-supplied pool, and every dispatcher call site passed the shared processing pool — whose permits some exec() callers already hold (process_bundle → reassembly/ delivery → exec). Once the pool saturated with permit-holding tasks all parked waiting for filter-task permits from the same semaphore, bundle processing deadlocked permanently; every shipped config has two Ingress readers at level 0, so the spawn path ran on every ingress. Remotely triggerable by withholding final fragments of N ADUs then releasing them together.

The FilterEngine now owns a dedicated unbounded TaskPool shared by all chains, and exec() no longer takes a pool parameter. Unbounded is safe: every spawned task is awaited within exec(), so concurrency is bounded transitively by the callers — and with no semaphore, the permit-holder-awaits-permit cycle cannot form. Regression test spawns exec() from callers holding all permits of a saturated BoundedTaskPool.

### DIFF
--- a/bpa/docs/filter_subsystem_design.md
+++ b/bpa/docs/filter_subsystem_design.md
@@ -212,9 +212,11 @@ The registry holds pre-built `FilterChain`s wrapped in an `Arc`. On each `exec` 
 
 This prevents writer starvation, is Send-safe for async runtimes, and avoids per-bundle rebuilds of the execution structure.
 
-### Rate-Limited Execution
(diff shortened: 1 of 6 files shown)


sylvain-pierrot (approved):
LGTM

Issue #603, opened 2026-07-08
hardy-bpv7: transitive dependency on std via time crate


felix-walter:
In 8b209465baaf1d7b544b194d8a89f58c86ef5f04, the `formatting` feature of the `time` crate was added, which [depends on `std`](https://github.com/time-rs/time/blob/main/time/Cargo.toml#L35), breaking `no_std` compatibility of `hardy-bpv7`. I tried removing that feature in a local copy of the `hardy-bpv7` crate and it still compiles - maybe it can be removed?

ricktaylor:
Hmmmm...  Let me have a look, as that's not what's wanted!

ricktaylor:
This is now fixed in v0.2.0

PR #611, opened 2026-07-09
fix(bpv7): drop time "formatting" feature to restore no_std; add no_std CI gate

ricktaylor:
time's `formatting` feature force-enables `time/std`, so bpv7 always linked std — silently breaking no_std for bare-metal targets. It was masked because `cargo build --no-default-features` on a std host still succeeds; the break only shows when building where std is absent. `formatting` was unused; drop it (keep `macros`, used by dtn_time's datetime!). The `std` feature already gates `time/std`.

Add a `no-std` CI job building cbor + bpv7 + eid-patterns --no-default-features for thumbv7em-none-eabihf — a bare-metal, 32-bit target — so an accidental std-pulling dependency feature fails in CI instead of shipping. bpv7's AtomicU64 needs the `critical-section` feature on Cortex-M (rlib build needs no impl). Compile-only: does not catch runtime `as usize` truncation.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,6 +46,49 @@ jobs:
       - name: Build all tests
         run: cargo test --no-run --locked --all-features --workspace
 
+  no-std:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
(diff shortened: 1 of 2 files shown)


ricktaylor:
Fixes #603

sylvain-pierrot (approved):
LGTM!

PR #612, opened 2026-07-09
fix(eid-patterns): correct scheme parsing, scheme/glob matching, and …

ricktaylor:
…union specificity

Four confirmed rescan findings, all in the published eid-patterns crate, each of which silently broke a documented pattern form in the RIB:

- R-11 parse.rs: parse_non_zero_decimal used exclusive '1'..'9', so a numeric scheme beginning with 9 (9:**, 91:**) failed to parse. Use the inclusive '1'..='9' per the %x31-39 grammar.
- R-12 lib.rs: EidPatternItem::matches fell through to `_ => false` for AnyNumericScheme/AnyTextScheme, so scheme-family wildcards matched nothing — including an Eid::Unknown decoded from the wire. Match against the EID's scheme (dtn=1, ipn=2, else the raw code).
- R-13 dtn_pattern.rs: do_glob matched against a double-slash target while patterns are built with a single slash, so single-star globs (dtn://node/*, dtn://*/svc) matched nothing under
require_literal_separator. Use a single slash; ** still spans separators.
- R-21 lib.rs: specificity_score returned None for multi-item unions, so a union route sorted as least-specific (behind *:**) and could not override broader routes. Score a union as its broadest (min) member.

Enables the deferred DTN glob-matching tests and adds regression tests.

### DIFF
--- a/eid-patterns/src/lib.rs
+++ b/eid-patterns/src/lib.rs
@@ -94,13 +94,14 @@ impl EidPattern {
 
     /// Harmonized Specificity Score.
     ///
-    /// Returns `None` for union sets (multiple items) or patterns violating
-    /// monotonic constraints.
+    /// A union set scores as its *broadest* (least specific) member, since it
+    /// matches the union of its members and is therefore only as narrow as the
+    /// widest one. Returns `None` for an empty set or if any member is
+    /// unscoreable.
     pub fn specificity_score(&self) -> Option<u32> {
         match self {
             EidPattern::Any => Some(0),
-            EidPattern::Set(items) if items.len() == 1 => items[0].specificity_score(),
-            EidPattern::Set(_) => None, // Union sets not valid for scoring
+            EidPattern::Set(items) => items.iter().map(|i| i.specificity_score()).min().flatten(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


eid-patterns/src/tests.rs
sylvain-pierrot, line 1: Moving it in a `eid-patterns/tests`??

eid-patterns/src/lib.rs
```diff
@@ -321,7 +344,12 @@ impl EidPatternItem {
             EidPatternItem::IpnPatternItem(i) => i.matches(eid),
             #[cfg(feature = "dtn-pat-item")]
             EidPatternItem::DtnPatternItem(i) => i.matches(eid),
-            _ => false,
+            // Scheme-family wildcards (`N:**` / `scheme:**`) match any EID of
+            // that scheme — including an `Eid::Unknown` decoded from the wire.
+            EidPatternItem::AnyNumericScheme(n) => eid_numeric_scheme(eid) == Some(*n),
+            EidPatternItem::AnyTextScheme(s) => {
+                eid_numeric_scheme(eid) == numeric_scheme_of_text(s)
```
sylvain-pierrot: If both helpers return None, we have `None == None` which return true… A fix would be:

```rust
EidPatternItem::AnyTextScheme(s) => {
      numeric_scheme_of_text(s).is_some_and(|n| eid_numeric_scheme(eid) ==
  Some(n))
  }
```

eid-patterns/src/tests.rs
ricktaylor: Good call - I wasn't planning a wholescale tidy up - but might be worth it for eid_patterns

eid-patterns/src/lib.rs
ricktaylor: I'm not sure that's any cleaner ;)

sylvain-pierrot: Having a None == None = true is a behavior expected?

eid-patterns/src/tests.rs
ricktaylor: Moved!

eid-patterns/src/lib.rs
ricktaylor: Yep - you're right - fix incoming

sylvain-pierrot (approved):
Perfect, LGTM

PR #613, opened 2026-07-09
fix(bpv7/editor): reject security blocks in remove_block

ricktaylor:
remove_block guarded against removing the primary/payload blocks and checked BIB coverage of the target, but — unlike push_block/insert_block/ update_block — never rejected the removed block itself being a BIB or BCB. Removing a BCB left its targets holding ciphertext with no covering BCB, surfacing ciphertext as plaintext on reparse; removing a BIB silently stripped integrity and left a dangling bib reference. Reject BlockIntegrity/BlockSecurity with Error::SecurityBlock; security blocks are managed only via Signer/Encryptor (remove_integrity/remove_encryption).


bpv7/src/bpsec/rfc9173/tests.rs
sylvain-pierrot, line 1: If this file you only public API, I would suggest to move it into `bpv7/tests`

ricktaylor: Let me investigate

ricktaylor: It's in the bpsec/rfc9173/tests so although testing public API it's behind the feature flag - so not too keen to move

ricktaylor: No I'm wrong - it's generic - let me move it

ricktaylor: Moved!

sylvain-pierrot (approved):
LGTM!

PR #614, opened 2026-07-09
fix(cbor): encode NaN as canonical half-precision f97e00

ricktaylor:
lossless_float_coerce rejects every narrowing of a NaN because the value-preservation check (into(f) == value) is always false for NaN, so f64::NAN emitted the 9-byte fb7ff8… form and f32::NAN the 5-byte fa7fc0… form — non-canonical output that the crate's own decoder flags as non-shortest. RFC 8949 §4.2.2 canonicalises NaN to the half-precision 0xf97e00, like the infinities in §4.2.1. Special-case is_nan() in the f64/f32 ToCbor impls to emit f97e00 for any NaN. Updates the two tests that asserted the old full-width bytes.

### DIFF
--- a/cbor/src/encode.rs
+++ b/cbor/src/encode.rs
@@ -652,7 +652,13 @@ impl ToCbor for f64 {
     type Result = ();
 
     fn to_cbor(&self, encoder: &mut Encoder) -> Self::Result {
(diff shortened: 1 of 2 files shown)


sylvain-pierrot (approved):
Approved!

PR #615, opened 2026-07-09
fix(bpv7/builder): don't panic in release when caller sets is_fragment

ricktaylor:
Builder::with_flags used a plain assert! on is_fragment — a release-build panic on a public-API call — even though the next line already normalises the flag to false. The fragment flag is owned by the fragmentation logic, not the caller. Use debug_assert! so misuse is still caught in debug builds while release normalises rather than crashes.


sylvain-pierrot (approved):
LGTM

PR #616, opened 2026-07-09
fix(bpv7/bpsec): accept RFC-permitted 8-16 byte BCB-AES-GCM IVs on de…

ricktaylor:
…crypt

decrypt only accepted a 12-byte IV (fixed-nonce Aes128Gcm/Aes256Gcm aliases), so any other length returned DecryptionFailed. RFC 9173 §4.3.1 permits any IV of 8-16 bytes (12 is only a SHOULD), so interoperable ciphertext from a peer using a non-12 IV could not be decrypted. aes-gcm parameterises the cipher by nonce size and implements both the 96-bit and GHASH-derived J0 constructions, so decrypt_middle dispatches on the wire IV length to the matching AesGcm<Aes, Un> (8..=16); decrypt_inner is already generic. Encrypt still emits 12-byte IVs. Parameters::from_cbor rejects IV lengths outside 8-16 (new Error::InvalidIvLength). Tests: 8/12/16 round trip and parse-time rejection of a 20-byte IV.

### DIFF
--- a/bpv7/src/bpsec/rfc9173/bcb_aes_gcm.rs
+++ b/bpv7/src/bpsec/rfc9173/bcb_aes_gcm.rs
@@ -69,8 +69,14 @@ impl Parameters {
             }
         }
 
+        // RFC 9173 §4.3.1: the IV length MUST be between 8 and 16 bytes.
+        let iv = iv.ok_or(Error::MissingContextParameter(1))?;
+        if !(8..=16).contains(&iv.len()) {
+            return Err(Error::InvalidIvLength(iv.len()));
+        }
+
         Ok(Self {
-            iv: iv.ok_or(Error::MissingContextParameter(1))?,
+            iv,
             variant: variant.unwrap_or_default(),
             key,
             flags: flags.unwrap_or_default(),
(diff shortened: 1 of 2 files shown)


bpv7/src/bpsec/rfc9173/bcb_aes_gcm.rs
```diff
@@ -397,22 +403,59 @@ impl Operation {
...
+        cek: &[u8],
+        aad: &[u8],
+        data: &[u8],
+    ) -> Result<zeroize::Zeroizing<Box<[u8]>>, Error>
+    where
+        Aes: aes_gcm::aes::cipher::BlockSizeUser<BlockSize = aes_gcm::aes::cipher::consts::U16>
+            + aes_gcm::aes::cipher::BlockCipherEncrypt
+            + KeyInit,
+    {
+        use aes_gcm::aes::cipher::consts::{U8, U9, U10, U11, U12, U13, U14, U15, U16};
```
sylvain-pierrot: Do we want to keep the import here?

ricktaylor: Uggh - not really - let me fix

ricktaylor: Fixed

sylvain-pierrot (approved):
LGTM

PR #617, opened 2026-07-09
fix(bpv7/bpsec): lock down OperationSet construction against empty sets

ricktaylor:
bib/bcb OperationSet had pub source/operations fields, so external code could build one with an empty operations map and hit the first().unwrap() panic in to_cbor (a public-API panic in published bpv7). Make the fields pub(crate) so only the crate (parser, Signer, Encryptor — all of which guarantee non-empty) can construct one, and add #[inline] pub source()/operations() accessors so external readers (the bundle inspect tool) can still inspect a parsed set. Harden the emit unwrap to an expect documenting the invariant.

### DIFF
--- a/bpv7/src/bpsec/bcb.rs
+++ b/bpv7/src/bpsec/bcb.rs
@@ -149,9 +149,10 @@ impl hardy_cbor::encode::ToCbor for OperationSet {
         // Targets
         encoder.emit(targets.as_slice());
 
-        // Context (an OperationSet is non-empty by construction)
+        // Context
         operations
             .first()
+            // SAFETY: An OperationSet is non-empty by construction
             .expect("OperationSet must contain at least one operation")
             .emit_context(encoder, &self.source);
 
--- a/bpv7/src/bpsec/bcb.rs
+++ b/bpv7/src/bpsec/bcb.rs
@@ -96,15 +96,32 @@ impl Operation {
 }
 
 /// A set of BCB operations sharing a common security source.
+///
+/// Fields are crate-private: an `OperationSet` is only ever produced by the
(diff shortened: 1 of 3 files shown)


bpv7/src/bpsec/bcb.rs
```diff
@@ -132,10 +149,10 @@ impl hardy_cbor::encode::ToCbor for OperationSet {
         // Targets
         encoder.emit(targets.as_slice());
 
-        // Context
+        // Context (an OperationSet is non-empty by construction)
         operations
             .first()
-            .unwrap()
+            .expect("OperationSet must contain at least one operation")
```
sylvain-pierrot, line 156: Safe?? Can you add a comment like `// SAFE: ….`?

bpv7/src/bpsec/bib.rs
```diff
@@ -100,10 +117,10 @@ impl hardy_cbor::encode::ToCbor for OperationSet {
         // Targets
         encoder.emit(targets.as_slice());
 
-        // Context
+        // Context (an OperationSet is non-empty by construction)
         operations
             .first()
-            .unwrap()
+            .expect("OperationSet must contain at least one operation")
```
sylvain-pierrot, line 124: Here as well? I’m wondering we should adapt this convention. What’s your take on this?

bpv7/src/bpsec/bcb.rs
ricktaylor: Yeah - it is safe, because it is asserted not to be possible - but yes

bpv7/src/bpsec/bib.rs
ricktaylor: yes, a //SAFETY: line is also really useful

sylvain-pierrot (approved):
LGTM

PR #618, opened 2026-07-09
fix(bpv7): reassemble indefinite-length byte-string block payloads fr…

ricktaylor:
…om the right offset

BlockWithNumber::from_cbor's ByteStream arm sliced &data[r] using chunk ranges that are relative to the value start (payload_start), so it read from the block header (type/number/flags/crc) instead of the payload chunks. A block whose block-type-specific data was encoded as a CBOR indefinite-length byte string — well-formed per RFC 8949 §4.1, reachable from any peer — was silently reassembled into corrupt bytes (emitted as a valid canonical bundle) or caused a valid bundle to be rejected.

Rebase each chunk range into the block buffer (payload_start + r), as the definite-length Bytes arm already does. Adds a regression test with an indefinite-length byte-string payload.

### DIFF
--- a/bpv7/src/block.rs
+++ b/bpv7/src/block.rs
@@ -437,7 +437,13 @@ impl hardy_cbor::decode::FromCbor for BlockWithNumber {
                                 ranges
                                     .into_iter()
                                     .fold(Vec::new(), |mut acc, r| {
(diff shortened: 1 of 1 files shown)


sylvain-pierrot (approved):
LGTM

PR #619, opened 2026-07-10
Memory-storage capacity watermarks, eviction fixes, and persistent-by-default server storage

ricktaylor:
Summary

The in-memory storage backends discarded bundles at capacity with metrics but no log signal, and bpa-server silently defaulted to those volatile backends. This PR adds edge-triggered high-watermark logging to both memory stores, fixes several eviction-correctness
bugs found along the way, and makes an unconfigured server default to persistent storage (SQLite metadata + local-disk bundles) with clear startup warnings whenever the volatile backends are explicitly chosen.

Capacity watermark logging

Both memory stores now emit a single info! line when occupancy crosses 95% of capacity and another when it falls back below 90%, with the exit line reporting everything evicted during the episode — so sustained pressure produces two log lines, not one per eviction.
The bundle store measures byte usage (the 95% line fires before any eviction, as eviction only starts at 100%); the metadata store measures live-entry count. Watermarks are computed subtractively (max - max/20) so the arithmetic cannot overflow usize on 32-bit
targets. A large overshoot that evicts straight through the hysteresis band reports both edges from the one call.

Eviction correctness fixes

- metadata_mem: tombstones were the most-protected entries. tombstone() promoted the fresh tombstone to MRU while long-Waiting live bundles sat first in line for eviction. Tombstones are now demoted to the LRU tail, so capacity evictions consume them before any live
bundle — and reaper-driven expiry (which tombstones through the dispatcher) now lands free slots exactly where eviction looks first.
- metadata_mem: silent eviction drift. The old put() path could evict without updating the gauges; every mutation now flows through one push-based funnel that distinguishes same-key replacement from capacity eviction, with gauges set from authoritative counts.
- metadata_mem: duplicate arrivals promoted tombstones. The presence check used get() (which reorders); now contains().
- metadata_mem: expired evictions are housekeeping, not data loss. With no background sweep in the store (the BPA reaper owns expiry), evicting an already-expired bundle is classified separately and excluded from the episode's data-loss tally.
- bundle_mem: replace() never evicted. Only save() ran the eviction loop, so a replace with larger data left usage above capacity until the next save. Both mutation sites now share one eviction path.
- bundle_mem: min_bundles = 0 allowed self-eviction. A save larger than the whole budget could evict its own just-stored bundle and return a dangling storage name; min_bundles is now clamped to ≥ 1.

New metrics: bpa.mem_metadata.evictions counter labelled kind = live | expired | tombstone.

Persistent-by-default server storage

bpa-server now defaults to SQLite metadata and local-disk bundle storage when no storage is configured (sqlite-storage joins the default feature set; cfg_select! falls back to memory for builds without the features). Both storage crates already ship platform-default
paths, so an unconfigured server just works. The memory backends must be selected explicitly and now warn! at startup that contents will not survive a restart.

Audited every BPA construction site in the workspace: the bpa crate's builder default remains memory by design (the persistent crates depend on bpa, so the flip lives in the server), which keeps the ping tool, benches, unit/pipeline tests, and fuzz harnesses on
memory; the bpa-server test scripts, interop suite, and compose config all pin their storage explicitly.

Documentation

User docs: storage configuration page now states the real defaults (and gained the memory backends' config-key tables); observability metrics tables split per backend, TCPCL-style, documenting the new counter and watermark lines. Design docs: storage subsystem doc
describes the actual eviction behaviour; bpa-server design doc's storage example fixed to the real field names. Rustdoc added across the touched public surface. A follow-up plan for common + bespoke storage backend metrics is recorded in bpa/docs/TODO.md.


bpa-server/src/config/storage.rs
```diff
@@ -25,11 +26,19 @@ pub enum MetadataStorageConfig {
...
     fn default() -> Self {
-        Self::Memory(Default::default())
+        cfg_select! {
+            feature = "sqlite-storage" => {
+                Self::Sqlite(Default::default())
+            }
+            _ => {
+                Self::Memory(Default::default())
+            }
+        }
```
sylvain-pierrot, line 41: I’m wondering if we remove the feature of “sqlite-storage” and add another feature “memory”. Like doing the other way around. What do you think?

bpa-server/src/config/storage.rs
```diff
@@ -47,7 +56,14 @@ pub enum BundleStorageConfig {
...
     fn default() -> Self {
-        Self::Memory(Default::default())
+        cfg_select! {
+            feature = "localdisk-storage" => {
+                Self::LocalDisk(Default::default())
+            }
+            _ => {
+                Self::Memory(Default::default())
+            }
+        }
```
sylvain-pierrot, line 76: Same for here, removing the feature "localdisk-storage” and adding “memory” feature

bpa-server/src/config/storage.rs
sylvain-pierrot: Memory should really be a particular use case

ricktaylor: Yeah - I'm not sure there's real benefit - the memory stuff is in the same crate - and there's no real reason to exclude it.  Features that removed dependencies make sense to me, conditional compilation - not so sure

sylvain-pierrot: My argument would be: without configuration a user would like to be as close as possible to a long running deployment, and memory is mainly for debug or corner use cases. 
A user should explicitly specify “memory” in Hardy config. So I would say, no "sqlite-storage” and "localdisk-storage” features, and they become the real default (without `cfg_select`)

ricktaylor: Have a look at the change - you either get the correct default, or a panic

sylvain-pierrot (approved):
Approved! This work was very necessary, thank you.

PR #625, opened 2026-07-13
fix(bpv7/builder): key the blocks map by wire block number

ricktaylor:
Builder::build inserted extension blocks into the returned Bundle.blocks map keyed by their enumeration index (0,1,2...) while numbering them on the wire as index+2. The first extension overwrote the primary (key 0) and the second collided with the payload (key 1), so a built bundle's blocks map disagreed with a parsed one, which every consumer keys by wire number. Key by the wire number instead. Adds a regression test.


sylvain-pierrot (approved):
LGTM!

PR #626, opened 2026-07-13
fix(bpv7): remove target CRC when signing the primary block (RFC 9173 §3.8.1)

ricktaylor:
BIB-HMAC-SHA2 signing removed the CRC from non-primary targets before generating the IPPT (RFC 9173 §3.8.1) but exempted the primary block, leaving its CRC in place. The IPPT was therefore computed over the CRC-bearing primary, producing a signature that would not interoperate with a conformant verifier (which computes the IPPT over the CRC-removed primary). RFC 9171 §4.3.1 permits the primary to carry no CRC when a BIB targets it.

Remove the primary's CRC on signing by re-emitting it canonically with no CRC and registering the bytes via set_canonical_primary, so the IPPT, the stored block, and the verifier all use the same form. The removal stays scoped to the BIB target: a primary included in the IPPT as scope context (§3.7 step 2) retains its CRC.

Also clear a removed target's BIB coverage in remove_from_bib_targets so rebuild_bundle never reports a dangling reference to a BIB that no longer exists.

Update the primary-signing test to assert the CRC is removed; add signer integration tests covering the target/context CRC cases and coverage clearing.

### DIFF
--- a/bpv7/src/bpsec/signer.rs
+++ b/bpv7/src/bpsec/signer.rs
@@ -172,12 +172,27 @@ impl<'a> Signer<'a> {
                     .blocks
                     .get(target)
                     .expect("Missing target block");
-                if *target != 0 && !matches!(target_block.crc_type, crc::CrcType::None) {
-                    editor = editor
-                        .update_block_inner(*target)
-                        .map_err(|(_, e)| e)?
-                        .with_crc_type(crc::CrcType::None)
-                        .rebuild();
+                if !matches!(target_block.crc_type, crc::CrcType::None) {
+                    // The rule applies to every target, including the primary
+                    // (RFC 9171 permits the primary to carry no CRC when a BIB
+                    // targets it).
+                    if *target == 0 {
+                        // Re-emit the primary canonically with no CRC and
+                        // register the bytes, so the IPPT (which reads the
+                        // primary via `block(0)`) and the rebuilt bundle agree
+                        // on the CRC-removed form.
+                        let mut primary = self.original.clone();
+                        primary.crc_type = crc::CrcType::None;
+                        let canonical = crate::bundle::primary_block::PrimaryBlock::emit(&primary)
+                            .map_err(editor::Error::from)?;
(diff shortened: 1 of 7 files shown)


bpv7/src/bpsec/signer.rs
```diff
@@ -172,12 +172,27 @@ impl<'a> Signer<'a> {
...
+                if !matches!(target_block.crc_type, crc::CrcType::None) {
+                    // The rule applies to every target, including the primary
+                    // (RFC 9171 permits the primary to carry no CRC when a BIB
+                    // targets it).
+                    if *target == 0 {
+                        // Re-emit the primary canonically with no CRC and
+                        // register the bytes, so the IPPT (which reads the
+                        // primary via `block(0)`) and the rebuilt bundle agree
+                        // on the CRC-removed form.
+                        let mut primary = self.original.clone();
```
sylvain-pierrot: Here you clone the whole `struct Bundle`. Would there be an alternative?

ricktaylor: Done

bpv7/tests/rfc9173.rs
sylvain-pierrot, line 1: Is this test file using the public API only?

ricktaylor: Yeah - I don't want to do too much refactoring right now, but mostly yes - but it's gated behind the rfc9173 feature...

sylvain-pierrot: Understood, if you want to take advantage of this PR for moving the file to `tests/` you can 🙂 
Otherwise LGTM!

ricktaylor: I'm moving it

sylvain-pierrot (dismissed):

bpv7/tests/rfc9173.rs
sylvain-pierrot: Great

sylvain-pierrot (approved):

PR #628, opened 2026-07-17
fix(bpa): don't set report_on_failure on blocks added to admin-record bundles

ricktaylor:
The forwarding path set the report_on_failure flag unconditionally on the PreviousNode, HopCount, and BundleAge blocks it adds to a transiting bundle. RFC 9171 sections 4.2.3-4/-5 forbid that flag on any block of a bundle that is an administrative record or has an anonymous source - the receiver has nowhere meaningful to report to - and our own ingress parser enforces exactly that, so a status report forwarded through a second Hardy node was rejected by it as InvalidFlags.

Set the flag only when the bundle's primary block permits reporting.

(ported from e2af3ad0 on refactor/parse, where the regression pipeline test lives)


sylvain-pierrot (approved):
LGTM

PR #629, opened 2026-07-17
fix(bpa): date the reassembled bundle from its earliest fragment arrival

ricktaylor:
poll_fragments already computed the minimum received_at across the fragment set but dropped it; ReassemblyResult::Done now carries it and the dispatcher stamps it on the re-entry metadata, so the ingress expiry check for no-clock sources (creation = received_at − age) is anchored to the first fragment's arrival rather than the trigger fragment's.

(cherry picked from commit a607f17f, docs hunk dropped)


sylvain-pierrot (approved):
LGTM

PR #630, opened 2026-07-17
fix(bpa): report bundle reception before the duplicate check

ricktaylor:
RFC 9171 section 5.6 reports on reception, and deduplication belongs to the later dispatch step, so a replayed or duplicate bundle is still reported as received - a sender may repeat a bundle deliberately, probing for status-report replies. Previously a refused duplicate was silently dropped before the reception report was generated.

This adopts the ordering already decided in the ingress status-report conformance notes (bpa/docs/TODO.md) and matches the refactor/parse ingress.

(ported from 9b59a5e3 on refactor/parse)


sylvain-pierrot (approved):
LGTM

PR #631, opened 2026-07-17
refactor(bpv7): use Arc for BPSec parameters so parse results are Send

ricktaylor:
The BIB/BCB OperationSet parameter maps were reference-counted with Rc, which made every type holding them - the operation sets, and any parse result carrying them - !Send. Callers that parse a bundle and then await (delivery, admin record handling) have to scope the parse into a block so the operation sets drop before the await point.

Swap Rc for alloc::sync::Arc. The clone/equality semantics are unchanged and the maps are immutable after parse, so the only cost is atomic reference counting, paid once per security block.

(ported from refactor/parse, where the streaming ingress holds the parser across awaits and needs this)

### DIFF
--- a/bpv7/src/bpsec/parse.rs
+++ b/bpv7/src/bpsec/parse.rs
@@ -1,6 +1,6 @@
 use super::*;
 use crate::error::HasInvalidField;
-use alloc::rc::Rc;
+use alloc::sync::Arc;
 use core::ops::Range;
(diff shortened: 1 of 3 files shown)


sylvain-pierrot (dismissed):
LGTM

sylvain-pierrot (approved):

PR #632, opened 2026-07-17
fix(rib): record adjacent neighbour as next-hop for via chains

ricktaylor:
A multi-level Via chain recorded the outermost via EID in metadata.read_only.next_hop: each enclosing level overwrote the deeper resolution with its own via, so egress filters saw an intermediate gateway rather than the neighbour the bundle is handed to. Carry each peer's resolved next-hop up unchanged so the adjacent neighbour EID from the Forward base case survives to the top.

Forwarding (the selected peer) was already correct; only the next_hop metadata was affected.

(cherry picked from commit 480d3bd26716926f3328661dd68258152722f3a9)

### DIFF
--- a/bpa/src/routing/rib.rs
+++ b/bpa/src/routing/rib.rs
@@ -575,6 +575,40 @@ mod tests {
     }
 
     #[test]
+    fn test_via_chain_next_hop_is_adjacent_neighbour() {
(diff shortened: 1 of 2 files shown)


sylvain-pierrot (approved):
LGTM

PR #633, opened 2026-07-17
Refactor/cbor parse tightening

ricktaylor:
Three related hardenings of bpv7's CBOR decode surface, extracted from the v0.3.0 refactor stack so they can be reviewed on their own:

- **`feat(cbor): add parse_exact`** — a decode helper for "exactly one item, no trailing bytes" that rejects anything after the decoded value.
- **`refactor(bpv7): route one-shot decodes through parse_canonical/parse_exact`** — adds a `parse_canonical` helper for leaf `FromCbor` impls whose wire form is a single bare value (block/CRC type codes, flags, lifetimes, DTN times, bundle age, BPSec context/scope/variant ids), decoding in canonical shortest form and rejecting non-shortest encodings, and adopts it across those impls in place of the hand-rolled decode-and-check patterns. Mechanical; no acceptance change where a check already existed.
- **`fix(bpv7): bounds-check wire-derived BPSec parameter/result ranges`** — the parameter/result ranges recorded in an `AbstractSyntaxBlock` were sliced into `source_data` unchecked, so a partial or mismatched buffer is a release-mode panic; every such slice now goes through `bounded_slice`, returning a clean `Error::SourceOutOfRange` instead. Defence in depth on main; the incoming streaming ingress makes partial buffers genuinely reachable.

### DIFF
--- a/bpv7/src/error.rs
+++ b/bpv7/src/error.rs
@@ -161,3 +161,25 @@ where
         Ok((t, true)) => Ok(t),
     }
 }
+
+/// Decode a `T` from the start of `data` in its canonical (shortest) form,
+/// returning the value and the bytes consumed. A non-canonical encoding —
+/// `T::from_cbor` reporting `shortest == false` — is rejected with
+/// `not_canonical`. The whole-slice counterpart of [`require_canonical`] (which
+/// decodes an array element), shared by leaf `FromCbor` impls whose wire form is
+/// a single bare value: block/CRC type codes, lifetimes, DTN times, BPSec
+/// context and variant ids. Generic over the caller's error so each keeps its
+/// own `NotCanonical`.
+pub(crate) fn parse_canonical<T, E>(data: &[u8], not_canonical: E) -> Result<(T, usize), E>
+where
+    T: hardy_cbor::decode::FromCbor,
+    T::Error: From<hardy_cbor::decode::Error>,
+    E: From<T::Error>,
+{
+    let (value, shortest, len) = hardy_cbor::decode::parse::<(T, bool, usize)>(data)?;
+    if shortest {
+        Ok((value, len))
+    } else {
+        Err(not_canonical)
+    }
+}
--- a/bpv7/src/block.rs
+++ b/bpv7/src/block.rs
@@ -108,11 +108,7 @@ impl hardy_cbor::decode::FromCbor for Flags {
     type Error = Error;
 
     fn from_cbor(data: &[u8]) -> Result<(Self, bool, usize), Self::Error> {
-        let (value, shortest, len) =
-            hardy_cbor::decode::parse::<(u64, bool, usize)>(data).map_err(Error::InvalidCBOR)?;
-        if !shortest {
-            return Err(Error::NotCanonical);
-        }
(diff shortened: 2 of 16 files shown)


bpv7/src/error.rs
```diff
@@ -161,3 +161,25 @@ where
...
+
+/// Decode a `T` from the start of `data` in its canonical (shortest) form,
+/// returning the value and the bytes consumed. A non-canonical encoding —
+/// `T::from_cbor` reporting `shortest == false` — is rejected with
+/// `not_canonical`. The whole-slice counterpart of [`require_canonical`] (which
+/// decodes an array element), shared by leaf `FromCbor` impls whose wire form is
+/// a single bare value: block/CRC type codes, lifetimes, DTN times, BPSec
+/// context and variant ids. Generic over the caller's error so each keeps its
+/// own `NotCanonical`.
+pub(crate) fn parse_canonical<T, E>(data: &[u8], not_canonical: E) -> Result<(T, usize), E>
```
sylvain-pierrot, line 173: Why are you passing the error as a parameter?
You always use `Error::NotCanonical`

sylvain-pierrot (commented):
Everything looks great! I have just one question for this PR

bpv7/src/error.rs
ricktaylor: Interesting....  let me investigate why I do this...

ricktaylor: Good question — it looks like a constant, but E really does change. The identifier Error::NotCanonical resolves to a different enum depending on the module: in dtn_time, lifetime, block, etc. it's the crate-root hardy_bpv7::Error::NotCanonical, while in bpsec and the RFC 9173 contexts it's bpsec::Error::NotCanonical — a separate leaf error type with its own variant. crc.rs is the giveaway: the CRC module's local Error has no NotCanonical variant, so its call site has to spell out crate::Error::NotCanonical explicitly.

Making the helper generic over E (rather than hard-coding one error type) lets every leaf parser keep its own focused error enum, per our "leaf error types, not crate-root Error everywhere" convention. The E: From<T::Error> bound also means the ? on the inner CBOR parse converts into the caller's error type directly, with no intermediate mapping. Passing the variant as a value is just the cheapest way to say "and this is your NotCanonical" — they're all unit variants, so constructing one eagerly costs nothing. The alternative (a NotCanonical-providing trait or a closure parameter) would be more machinery for the same six call sites.

sylvain-pierrot: If the caller knows the error, he cannot use a map_error? 
It looks a bit unusual to select the error to return as a parameter. 🤔

ricktaylor: So are you suggesting `pub(crate) fn parse_canonical<T>(data: &[u8]) -> Result<(T, usize), ()>`

And calling `parse_canonical::<T>(data).map_err(|_| Error::NonCanonical)?`

vs:

`parse_canonical::<T>(data,Error::NonCanonical)?`

I'm not sure that's better?

sylvain-pierrot: I’m fine with your solution :)

sylvain-pierrot (approved):
LGTM!!

PR #634, opened 2026-07-18
TCPCLv4: remove head-of-line blocking from the session protocol loop

ricktaylor:
Two changes with a shared goal: the session's protocol loop never waits on anything slower than the socket, and the connection pool actually delivers the parallelism it documents.

### Session rework (`0740eb41`)

The session reader previously awaited `sink.dispatch()` inline and a per-segment oneshot through the writer task. A storage write therefore stalled segment reception, acknowledgment processing, and outbound
segment flow — and every remote sender's transfer cycle included our ingest latency. Two peers pushing large transfers simultaneously could also deadlock (both blocked writing, neither reading, keepalives
stopped).

The session now runs three cooperating tasks:

- A per-session **ingest task** owns dispatch to the BPA and emits every XFER_ACK in segment-arrival order from a bounded FIFO. The final ack of a transfer is still only sent after dispatch returns — the
responsibility-transfer semantics that let the sending BPA delete on `Sent` are unchanged and now documented in `docs/design.md`. A failed dispatch leaves the transfer unacknowledged so the peer retransmits; BPA
dedup absorbs duplicates.
- The **writer task** gains a fire-and-forget `Feed` command (no per-segment round trip), flushes when its command queue runs dry, and exposes a cancel-safe `reserve()`/permit API the session uses to write while
still polling the reader. Backpressure reaches the peer through the channel bounds, never by stalling the protocol loop.

Also fixes a latent bug: a mid-transfer XFER_REFUSE now clears all outstanding ack expectations for the refused transfer (RFC 9174 §5.2.2 sends no further acks for it); stale expectations previously
desynchronised the strict-FIFO ack matcher and tore the session down.

### Connection pool ordering (`d9e796ce`)

`try_send` implemented queue-behind-busy-session *before* the capacity check, so a new connection was only dialed when the pool was empty and the documented burst capacity never engaged. It now checks capacity
between the idle and queue steps and signals a dial when there is room. Queueing on a busy session remains as the at-capacity path and — via a no-dial fallback pass in `Cla::forward` — when the peer cannot
accept new connections (asymmetric reachability, RFC 9174 §3.3).

### Wire compatibility and testing

No wire-format change: same segments, same acks, same order; only timing differs (acks no longer gated behind storage for non-final segments, flush-on-idle). Unit tests pin ingest ack ordering, dispatch-failure
behaviour, and all four pool paths. Full interop run against the TCPCLv4-speaking peers exercised the ping chain end to end.

Known deliberate deferrals (recorded in `docs/design.md` Future Work): outbound transfer windowing awaits the egress-policy work; priority-queue-to-session mapping; passive-side writer task tracking.


sylvain-pierrot (approved):
LGTM. Good job :)

PR #641, opened 2026-07-23
perf(sqlite-storage): run WAL connections at synchronous=NORMAL

ricktaylor:
synchronous is a per-connection pragma (unlike the persistent WAL journal mode), so it is set in the same per-connection batch at both connection sites: the pool's new_connection and the initial open.

At the SQLite default of FULL, every commit fsyncs the WAL, and the metadata store commits on each bundle status transition — on fsync-expensive storage (network-attached disks in particular) those per-commit syncs dominate the per-bundle forwarding path. NORMAL fsyncs at checkpoint instead: a crash preserves consistency and loses at most the un-checkpointed tail of commits, which restart recovery already tolerates — bundle data storage is ground truth, and data whose metadata is missing is re-ingested at startup.

### DIFF
--- a/sqlite-storage/CHANGELOG.md
+++ b/sqlite-storage/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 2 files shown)


sylvain-pierrot (approved):
LGTM

PR #642, opened 2026-07-24
Refactor/tcpclv4

### LINKED ISSUE #292: Add mutual TLS support for TCPCLv4
Add mutual TLS (mTLS) support to TCPCLv4 connections, so servers can require client certificates and clients can send theirs. Some `TODO(mTLS)` comments related to mTLS have also been added to the code, introduced in PR #291.

### LINKED ISSUE #411: Migrate from rustls-pemfile to rustls-pki-types
### Context

`cargo audit` reports `rustls-pemfile` v2.2.0 as unmaintained ([RUSTSEC-2025-0134](https://rustsec.org/advisories/RUSTSEC-2025-0134)).

The crate repository was archived in August 2025. The latest version of `rustls-pemfile` is already a thin wrapper around the PEM parsing code included in `rustls-pki-types` since 1.9.0.

### Dependency chain

```
rustls-pemfile v2.2.0
└── hardy-tcpclv4
    ├── hardy-bpa-server
    ├── hardy-tcpclv4-server
    └── hardy-tools
```

### Fix

Replace the `rustls-pemfile` dependency in `hardy-tcpclv4` with `rustls-pki-types` >= 1.9.0 and migrate call sites to the [`PemObject`](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/pem/trait.PemObject.html) trait.

### Notes

- No security vulnerability is associated with this advisory — it is informational only
- The migration should be straightforward as `rustls-pemfile` is already a thin wrapper over the same code

sylvain-pierrot:
# Rebuild TCPCLv4 around an entity, fluent builders, and a server-owned configuration schema

> Closes the following issues:
> - Fixes #292: mutual TLS is fully implemented, with the `client-auth: off | optional | required` policy, the node identity presented as the client certificate when dialing, and real-handshake tests covering all three policies plus the rogue-CA cases; every `TODO(mTLS)` the issue references is gone.
> - Fixes #411: the unmaintained `rustls-pemfile` dependency (RUSTSEC-2025-0134) is removed; PEM certificates and keys are loaded through `rustls::pki_types::pem::PemObject`.

## What this changes and why

Previously, the `hardy-tcpclv4` library parsed its own configuration file format, and TLS behavior was spread across loosely coupled options whose invalid combinations were caught (or missed) by scattered runtime checks. This PR inverts that: **the library now exposes a programmatic API with its invariants in the type system, and the configuration file format is owned by the applications that have config files** (`hardy-tcpclv4-server`, and `bpa-server`'s `clas` entries, kept in lockstep).

## The library

The CLA is now a `Tcpclv4` entity constructed by `Tcpclv4::builder()`. Builder inputs carry their own rules as types: `ContactTimeout` bounds the contact-header wait to RFC 9174's 1 to 60 seconds, `KeepaliveInterval` carries the interval with zero as the wire's own disabled encoding (RFC 9174 Section 4.7), and the MRUs, connection rate limit, and per-peer outstanding-transfer bound are `NonZero` integers. Listeners are explicit per RFC 9174's "zero or more listening elements" (no call means dial-only, `listen_default()` is the IANA-registered `[::]:4556`), and sockets are bound eagerly at build so port conflicts fail startup rather than a background task. A second registration of an already-registered entity is refused rather than silently swallowing the new sink.

TLS material is its own product: `tls::Tls::builder()` chains a mandatory trust anchor, an optional identity (certificate and key, only ever as a pair), and whether plaintext sessions are refused, then loads and validates everything in one `build()`. Reaching the insecure trust policy requires spelling out `.dangerous().insecure_skip_verify()`, and every certificate error names the file it is about.

**Mutual TLS is new**: `client-auth` (`off` | `optional` | `required`) has the listener request and verify dialers' certificates against `ca-certs`, with `optional` as the fleet-migration step, and the single identity doubles as the client certificate when dialing into a mutual-TLS peer. It is proven by real handshake tests over an in-memory duplex with a throwaway PKI, covering all three policies plus rogue-CA and untrusted-server cases.

One correctness fix fell out of the review: the contact header's `CAN_TLS` flag is now honest per role. A listener whose TLS material had no identity used to advertise TLS it couldn't serve, and a peer accepting the offer could crash the process.

## The errors

The crate's errors are restructured as a hierarchy. `error.rs` holds the generic, non-exhaustive `Error` for everything TCPCLv4: entity assembly (`BindListener`, `RequiredTlsWithoutIdentity`), registration (`AlreadyRegistered`), and a wrapper variant per sub-concept (`Tls(tls::Error)`, `Session(session::Error)`). `session::Error` owns how a session, or the establishment of one, ended, with variants named for what happened (`PeerHangup`, `PeerTimeout`, `ShuttingDown`, `PeerTerminated`, `LocalShutdown`, ...), the wire dialect's `codec::Error` wrapped as its `Codec` variant, and I/O reclassified from dialect to transport at conversion. `transport::Error` and `builder::Error` are gone, the session teardown dispatch and its metric labels are exhaustive matches again, and `Tcpclv4::connect` surfaces handshake failures through `Error::Session`.

## The configuration surface

Listeners and keepalives get explicit spellings for their "off" states, so no capability hides behind a YAML `null` that TOML and environment variables cannot express:

| Before | After |
|---|---|
| `address` (one listener; `null` meant dial-only) | `listeners`, a list with one entry per listening element: absent is the registered default, `[]` is dial-only, and multiple listeners are now reachable from config |
| `keepalive-interval: null` or `0` disabled | `0` disables; an explicit `null` is refused at parse with the replacement named |

Scalar invariants are carried by the schema types (`NonZero` integers and a `ContactTimeout` adapter), so zero and out-of-range values are rejected when the file is parsed, before any socket is touched. The `max-outstanding-transfers` bound from the deferred transfer-outcome work is exposed end-to-end (schema, examples, and user docs).

The `tls` section is restructured so that mistakes are hard to write and the dangerous option is impossible to miss:

| Before | After |
|---|---|
| `require-tls: true` (entry level) | `tls.required: true` (can't require TLS without configuring it) |
| `cert-file` / `private-key-file` | `identity.cert-file` / `identity.key-file` (a lone half is a parse error; `private-key-file` still accepted as an alias) |
| `debug.accept-self-signed: true` | `insecure-skip-verify: true` (deliberately loud; overrides `ca-certs` with a startup warning instead of silently ignoring it, so a debug session is one line to flip) |
| *(no equivalent)* | `client-auth: off \| optional \| required` (new: mutual TLS) |

A trust anchor is mandatory, and an empty or missing CA directory is a startup error.

## The servers

`hardy-tcpclv4-server` gains a small `Tcpclv4Server` type owning the full lifecycle (build, register with the BPA, dial static peers, graceful unregister on shutdown), behind a `main` that only loads config and runs it. Its config mapping, and bpa-server's, are 1:1 transcriptions of the file onto the builders; all judgment lives in the library or at parse, so both surfaces produce identical errors in the config's own vocabulary.

## Compatibility

Config-file changes are breaking and documented in each crate's CHANGELOG; the user documentation, example configs, and test docs are updated to match. The library API is breaking (pre-1.0). Test coverage grew alongside: builder defaults and invariants, the keepalive negotiation, TLS trust rules including the insecure override, real mutual-TLS handshakes, and the parse-level rejections (invalid scalars, `null` spellings, dial-only listeners) at both config surfaces.


ricktaylor (approved):
All good!

PR #643, opened 2026-07-24
Feat/cla transfer outcome

ricktaylor:
Adds deferred CLA transfer outcomes to the BPA↔CLA contract, and adopts them in tcpclv4.

- bpa: forward may answer Accepted — the CLA takes ownership and reports Delivered/Failed later via the new Sink::transfer_outcome, keyed by bundle ID. Accepted bundles persist as BundleStatus::ForwardAckPending until the outcome, peer removal, restart, or expiry resolves them; a deferred Failed re-enters dispatch per-bundle. Design rationale is in bpa/docs/design.md § Deferred CLA Transfer Outcomes.
- proto: the wire mirror — ForwardBundleRequest.bundle_id, an accepted result variant, and TransferOutcomeRequest. No capability negotiation: deferral is a per-bundle choice.
- tcpclv4: forwards resolve out-of-band; transfers overlap across pooled connections instead of serialising behind a held call, with max-outstanding-transfers (per peer, default 16) as the admission bound and withheld verdicts as flow control back to the BPA.


sylvain-pierrot (approved):
Amazing work! LGTM

PR #645, opened 2026-07-28
perf(bpa): avoid heap allocation in CBOR precheck error path

g4dpz:
Replace BpvError/CborError (which allocate String per error) with a dedicated PrecheckError enum using &'static str via Display. Under adversarial traffic, every malformed packet no longer triggers a heap allocation.


ricktaylor (approved):
This will be subsumed in later work - but definitely worth having now!  Thanks

ricktaylor:
Can you sign your commits please?

g4dpz:
Hi,

GPG or CLA ?

- Dave

PR #647, opened 2026-07-28
perf(bpa): use SmallVec in filter chain to avoid per-bundle heap allo…

g4dpz:
…cation

Replace Vec with SmallVec<[_; 4]> for the handles and results collections in Level::run_readers(). Most filter levels have ≤4 readers, so the common case stays entirely on the stack.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1842,6 +1842,7 @@ dependencies = [
  "metrics",
  "rand 0.10.2",
  "serde",
+ "smallvec",
  "thiserror",
  "time",
  "tokio",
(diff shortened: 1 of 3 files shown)


ricktaylor:
Hi David,

I'm not going to accept this - not because it's wrong, but for v0.3.0 we are doing a wholesale rewrite of the filter subsystem to make it much leaner all round.

Very happy to ensure we do use SmallVec in these kinds of places in the future.

ricktaylor:
Thanks David — the idea is right, but I'm going to close this one rather than merge it: the filter engine and chain are being replaced wholesale by the filter-subsystem redesign that follows the streaming refactor (registration freezes into plain slices at build time, and these two files go away). Merging a tweak to code with that short a shelf life isn't a good use of your rebase effort or review time.

The SmallVec instinct itself is sound though — if the new engine ends up with a per-bundle spawn path, the same trick applies there, and I'd welcome it against the new code once it lands. Meanwhile the #653 "safe now" group is still open if you fancy a quick win.

PR #648, opened 2026-07-28
perf(cbor): single extend_from_slice in emit_uint_minor

g4dpz:
Write the marker byte and value bytes into a stack-local [u8; 9] buffer, then emit the entire encoded head with one extend_from_slice call. This eliminates multiple push/extend calls per integer emission — each of which individually checks capacity and may trigger reallocation — and gives the compiler a single memcpy to optimise.

emit_uint_minor is called for every integer, length prefix, and tag in the CBOR output, making this the highest-frequency function in the encoder.

### DIFF
--- a/cbor/src/encode.rs
+++ b/cbor/src/encode.rs
@@ -132,29 +132,42 @@ impl Encoder {
     }
 
     fn emit_uint_minor(&mut self, major: u8, val: u64) {
-        const U8_MAX: u64 = (u8::MAX as u64) + 1;
-        const U16_MAX: u64 = (u16::MAX as u64) + 1;
-        const U32_MAX: u64 = (u32::MAX as u64) + 1;
-
-        match val {
-            0..24 => self.data.push((major << 5) | (val as u8)),
-            24..U8_MAX => {
-                self.data.push((major << 5) | 24u8);
-                self.data.push(val as u8)
+        // Write the marker byte and value bytes into a stack buffer, then
+        // extend once — avoids multiple push/extend calls and the
+        // associated per-call capacity checks.
+        let mut buf = [0u8; 9];
+        let head = major << 5;
+        let len = match val {
+            0..24 => {
+                buf[0] = head | (val as u8);
+                1
             }
-            U8_MAX..U16_MAX => {
-                self.data.push((major << 5) | 25u8);
-                self.data.extend((val as u16).to_be_bytes())
+            24..=0xFF => {
+                buf[0] = head | 24;
+                buf[1] = val as u8;
+                2
             }
-            U16_MAX..U32_MAX => {
-                self.data.push((major << 5) | 26u8);
-                self.data.extend((val as u32).to_be_bytes())
+            0x100..=0xFFFF => {
+                buf[0] = head | 25;
+                let bytes = (val as u16).to_be_bytes();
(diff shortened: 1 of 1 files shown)


ricktaylor:
Hi David,

Thanks for this!

Having done some further research - the true zero copy approach is to call reserve() on the Vec once we know the ultimate size, and then `extend_from_slice(v.to_be_bytes())` as LLVM is smart enough to avoid the unnecessary bounds check, and we avoid the static bounce buffer

ricktaylor:
Following up on my earlier comment — having looked at the codegen trade-offs more closely: the bounce buffer collapses two capacity checks into one, but &buf[..len] turns the append into a variable-length memcpy, losing the fixed-width stores the current per-arm to_be_bytes() appends get. My reserve-then-extend_from_slice suggestion keeps the fixed-size copies with a never-taken grow branch, which I'd expect to be at least as good — but honestly all three variants are likely within noise next to the block-data copies that dominate encoding, especially now #651 has removed the reallocation cost.

So before we churn a hot, stable function on instinct: could you add a small criterion bench for Encoder (there isn't one yet — it'd be a genuinely useful contribution beyond this PR) and show the delta? If it's measurable, I'll happily take the winning variant; if it's noise, we keep the simplest code.

g4dpz:
Hi Rick,

Ive added a criterion benchmark for the Encoder (on branch `feat/encoder-benchmark`, happy to submit as a separate PR). Here are the baseline results against your current main (with #651 merged but without the bounce-buffer change from this PR):

| Benchmark | Time |
|-----------|------|
| `encode/uint/u8_inline` (1 byte) | 25.6 ns |
| `encode/uint/u8_extended` (2 bytes) | 26.1 ns |
| `encode/uint/u16` (3 bytes) | 25.9 ns |
| `encode/uint/u32` (5 bytes) | 25.6 ns |
| `encode/uint/u64` (9 bytes) | 49.1 ns |
| `encode/mixed_integers_10` | 138 ns |
| `encode/array_10_elements` | 143 ns |
| `encode/bytes/1KB` | 87 ns |
| `encode/bundle_like_1KB` | 207 ns |
| `encode/capacity/new` (10 ints) | 63 ns |
| `encode/capacity/with_capacity` (10 ints) | 39 ns |

**Observations:**

1. `with_capacity` (#651) delivers a clear 38% improvement for small encodings — that was the real win
2. `emit_uint_minor` calls are 25–26ns for u8–u32 (indistinguishable from each other) and ~49ns for u64 — the difference is the 8-byte write itself, not the capacity check path
3. A 1KB bundle encoding takes 207ns, dominated by the byte-string memcpy (~87ns for the data alone vs ~140ns for the 10 integer headers combined)

**Conclusion:**

You were right — now that #651 has removed the reallocation cost, the three `emit_uint_minor` variants (current, bounce buffer, reserve-then-extend) are likely within noise relative to the block-data copies that dominate real encoding. I think the simplest code wins here.

Happy to close this PR and submit the benchmark as a standalone contribution instead. Let me know your preference.

ricktaylor:
Can we close this one - I will happilly accept a benchmarking PR!

g4dpz:
Closing as agreed — the emit_uint_minor variants are in the noise now that #651 removed the reallocation cost. Will submit the benchmark as a standalone PR.

PR #649, opened 2026-07-29
perf(bpa): use SmallVec in route table lookup

g4dpz:
Replace `Vec<(u32, &Eid)>` with `SmallVec<[(u32, &Eid); 4]>` in `find_recurse()` and `select_peer()`.

Most routes resolve to 1–4 ECMP peers, so the common case stays entirely on the stack — no heap allocation per bundle dispatch.

**Note:** `SmallVec`'s `Drop` impl extends borrow scopes beyond what `Vec` requires. To satisfy the borrow checker, `Rib::find()` is restructured into `find()` + `dispatch_from_lookup()` with a destination clone to break the `&bundle` ↔ `LookupResult<'_>` lifetime cycle. The clone is a single `Eid` (typically a few bytes for IPN) and happens once per dispatch — far cheaper than the heap allocation it replaces.

**Changes:**
- `bpa/src/routing/table.rs`: `SmallVec<4>` for peers vec + `sorted_insert`
- `bpa/src/routing/rib.rs`: restructured `find()`, updated `select_peer()` signature

**Testing:**
- `cargo test -p hardy-bpa --all-features` — 112 tests pass
- `cargo clippy -p hardy-bpa --all-features -- -D warnings` clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1842,6 +1842,7 @@ dependencies = [
  "metrics",
  "rand 0.10.2",
  "serde",
+ "smallvec",
  "thiserror",
  "time",
  "tokio",
--- a/bpa/Cargo.toml
+++ b/bpa/Cargo.toml
@@ -53,6 +53,7 @@ foldhash = { version = "0.2", default-features = false }
 lru = { version = "0.18", default-features = false }
 thiserror = { version = "2", default-features = false }
 arc-swap = "1"
+smallvec = "1"
 futures = { version = "0.3", default-features = false, features = ["alloc", "async-await"] }
 serde = { version = "1", default-features = false, features = ["derive","rc","alloc"], optional = true }
 base64 = { version = "0.22", default-features = false, features = ["alloc"], optional = true }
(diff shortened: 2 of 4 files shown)


ricktaylor:
Thanks David — closing this one too, for the same shelf-life reason as #647 plus two of its own:

* The route table is queued for its own redesign (companion to the filter-subsystem work that follows the streaming refactor), so find_recurse and the LookupResult shape don't survive either.
* As written, the borrow-scope restructuring adds a bundle.bundle.destination.clone() on every lookup — an unconditional Eid clone (heap-backed for dtn-scheme EIDs) to enable a SmallVec that only helps when ECMP actually fans out. On the common single-route path that's likely a net pessimisation, which is worth remembering as a pattern: a lifetime workaround that clones on the hot path can cost more than the allocation it saves.
* It also needs a rebase over #632, which landed today and reworks the same next-hop code.

Same offer as #647 stands: once the routing redesign lands, bounded-fanout lookups are exactly where SmallVec thinking belongs, and I'd welcome it against the new code.

Issue #653, opened 2026-07-30
perf: Use Vec::with_capacity where size is known or estimable


g4dpz:
## Summary

A scan of the codebase found ~32 instances of `Vec::new()` in non-test code where the capacity is known or reasonably estimable at construction time. Switching to `Vec::with_capacity()` would eliminate unnecessary reallocations.

## High-impact candidates (hot paths or known sizes)

| File | Line | Context | Suggested fix |
|------|------|---------|---------------|
| `bpa/src/filter/engine.rs` | 189 | `handles` followed by `for _ in 0..2` | `with_capacity(2)` |
| `bpa/src/storage/bundle_mem.rs` | 228 | `snapshot` built from all shards | `with_capacity(total_entries)` |
| `bpv7/src/editor.rs` | 198–199 | `unchanged` / `new_chunks` from extensions iterator | `with_capacity(extensions.len())` |
| `bpv7/src/editor.rs` | 1214, 1331 | `ext_chunks` from block iteration | `with_capacity(blocks.len())` |
| `bpv7/src/cmp.rs` | 68 | `diffs` from block comparison | Estimate from block count |
| `tcpclv4/src/session.rs` | 1151 | `acks` from known segment list | `with_capacity(segment_count)` |
| `tcpclv4/src/codec.rs` | 148, 652 | `session_extensions` / `transfer_extensions` | `with_capacity(2)` (usually 0–2) |
| `proto/src/client/cla.rs` | 216 | `try_fold(Vec::new(), ...)` on node_ids | `with_capacity(node_ids.len())` |
| `proto/src/client/routing.rs` | 139 | Same pattern | `with_capacity(node_ids.len())` |

## Lower priority (init-time / cold paths)

- `bpa/src/routing/rib.rs:71` — agents vec (populated at startup)
- `bpa/src/policy/htb_policy.rs:42` — ranges vec (config parsing)
- `bpv7/src/builder.rs:57` — extensions (built incrementally)
- `bpa/src/node_ids.rs:212` — endpoints (startup)

## Notes

The codebase already uses `SmallVec` effectively in performance-critical spots (bpsec, routing table, bundle parsing). The items above are the remaining cases where a simple `with_capacity` would avoid heap churn on the hot path.

Happy to submit a PR for these.

ricktaylor:
David,

[...]

PR #655, opened 2026-08-07
perf: pre-size Vecs where capacity is known (#653 safe-now group)

g4dpz:
The "safe now" group from #653 — conflict-free with the in-flight streaming refactor.

**Changes:**

1. **`proto/src/client/routing.rs`** — capture `node_ids.len()` before `into_iter()`, use `Vec::with_capacity(count)` in the try_fold
2. **`bpa/src/node_ids.rs`** — use `seq.size_hint()` in serde `visit_seq` to pre-allocate the endpoints vec
3. **`bpa/src/routing/rib.rs`** — pre-size the agents vec in `shutdown_agents()` from the map length before draining

**Testing:**
- `cargo fmt --check` clean
- `cargo clippy --all-features --all-targets -- -D warnings` clean
- `cargo test -p hardy-bpa -p hardy-proto --all-features` passes

### DIFF
--- a/bpa/src/node_ids.rs
+++ b/bpa/src/node_ids.rs
@@ -209,7 +209,7 @@ impl<'de> serde::Deserialize<'de> for NodeIds {
             where
                 A: serde::de::SeqAccess<'de>,
             {
-                let mut endpoints = Vec::new();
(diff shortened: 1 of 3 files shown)


ricktaylor (approved):
LGTM!

PR #656, opened 2026-08-07
perf(bpa-server): eliminate Vec allocation in BPSec key lookup

g4dpz:
Replace the intermediate `Vec<&Operation>` with iterator-based logic in `PatternKeySource::key()`.

The served-operations check and the key_ops matching now use `.any()` directly, avoiding a heap allocation on every BPSec block lookup.

**Changes:**
- `bpa-server/src/bpsec.rs`: replace `collect::<Vec<_>>()` with chained `.any()` iterators

**Testing:**
- `cargo fmt --check` clean
- `cargo clippy -p hardy-bpa-server --all-features -- -D warnings` clean
- `cargo test -p hardy-bpa-server --all-features` — 62 tests pass


ricktaylor (approved):
Good catch!

PR #664, opened 2026-08-11
Refactor/config ownership

sylvain-pierrot:
# Configuration is pure data: servers own the schemas, libraries own typed construction

> Stacked on the TCPCLv4 rebuild PR; this branch extends the same ownership line across the whole workspace.

## What this changes and why

Previously, configuration concerns were smeared across every layer: library crates (`hardy-bpa` and all four storage backends) shipped serde `Config` structs, defaults were restated at each level above the one that used them, and bpa-server's startup was spread between `main`, free functions, and `build()` methods hidden inside config modules. This PR draws one line through all of it: **config modules are pure data owned by the server crates, and libraries expose typed constructors and builders that own their defaults privately, at the point of use.**

## The server

bpa-server gains a `BpaServer` type owning the whole lifecycle, behind a `main` that is now a pure composition root: clap argument parsing (replacing getopts), tracing, and the task pool with signal wiring. `BpaServer::new` assembles every subsystem inline, storage backends, filters, BPSec keys, static routes, built-in services, policies, and CLAs, so startup reads as one top-to-bottom recipe, and every config-derived error names its subject (`CLA 'x': segment-mru must be greater than zero`). `run` starts the BPA, serves the gRPC front end, waits for cancellation, and shuts down in dependency order.

The `config/` directory is reduced to pure serde types with unique, self-describing names (`StorageConfig`, `BPSecConfig`, `ClaConfig`, `Tcpclv4Config`, …): the tcpclv4 schema folds into `cla.rs`, the one-type policy and static-routes modules fold into `mod.rs`, and nothing in `config/` can build, load, or validate anything beyond what serde expresses.

## The libraries

Every library that previously deserialized its own configuration now exposes construction the tcpclv4 way:

| Crate | Before | After |
|---|---|---|
| `hardy-bpa` | `Config` structs for the memory storages and the RFC 9171 filter | `MetadataMemStorage::new` / `BundleMemStorage::new` with `Option` knobs and dimension-named parameters; `Rfc9171ValidityFilter` with fluent setters |
| `hardy-sqlite-storage` | `Config` + free `new()` | `SqliteStorage::new(db_dir, db_name, upgrade)` |
| `hardy-localdisk-storage` | `Config` + free `new()` (+ `serde` feature) | `LocalDiskStorage::new(store_dir, fsync, upgrade)` |
| `hardy-postgres-storage` | `Config` + free `new()` | `PostgresStorage::builder()`: `Duration` timeouts, `NonZeroU32` pool knobs |
| `hardy-s3-storage` | `Config` + free `new()` (+ `serde` feature) | `S3Storage::builder(bucket)` with the `PartSize` newtype |

Invalid states became unrepresentable along the way: the S3 5 MiB part-size minimum lives in `PartSize`, a zero-connection postgres pool and a zero `min-bundles` floor cannot be expressed, and a multipart threshold below the part size is rejected at build (previously an unenforced doc claim). The bundle cache owns its LRU defaults (`bundle_cache.rs`), `BpaBuilder` carries unset knobs as `None` and caches any configured bundle storage unless `no_cache()`, and it is constructible only through `Bpa::builder()`.

## The configuration surface

The file keys are unchanged; what changed is honesty at the edges. Absent keys now defer to the library's default instead of restating it, so a config dump shows what the operator wrote rather than a frozen copy of internals. Values that were silently corrected or ignored are now errors: `min-bundles: 0` and `max-connections: 0` fail at parse, a missing S3 `bucket` fails at parse instead of startup, and the threshold/part-size rule is enforced. The storage reference documentation is corrected to match, including a wrong key name (`endpoint-url`, not `endpoint`) and four undocumented S3 keys.

## Compatibility

All library API changes are breaking (pre-1.0) and recorded per crate in the CHANGELOGs, alongside the bpa-server config-behavior notes. Test coverage moved with the code: the BPSec loading tests live with `PatternKeySource::load`, the storage suites cover the new constructors, and all gates pass across full, partial, and minimal feature sets.


ricktaylor (approved):
LGTM!

PR #665, opened 2026-08-12
fix(bpa,bpa-server): align node identity with the RFC admin-endpoint model

sylvain-pierrot:
# Name the node's identity for what it is: administrative endpoints

## What this changes and why

RFC 9171 and RFC 9758 give a node exactly one administrative endpoint per scheme (`ipn:[A.]B.0`, `dtn://node/`), and every other EID that identifies the node derives from it. Hardy's implementation already enforced this model precisely: the config accepted only admin-endpoint forms, at most one per scheme, and computed everything else. But the vocabulary said otherwise: the key was called `node-ids`, describing the derived superset rather than the stored identity. This PR aligns the names and the two remaining edges with the RFC model: **configuration stores administrative endpoints; node IDs are a computed relation; and the admin endpoint's service ids cannot be claimed.**

## The configuration surface

The `node-ids` key is renamed `admin-endpoints`, with `node-ids` kept as an alias so existing configs and scripts keep working (the interop test fixtures deliberately stay on the alias, giving it end-to-end CI coverage). The parse behaviour is unchanged and was already strict: only admin-endpoint forms are accepted, at most one per scheme, LocalNode rejected. The user documentation now says what the key is: the node's administrative endpoints, from which all other node IDs derive.

## The library

Registering a service at an admin endpoint's service id (ipn service 0, or the zero-length dtn demux) is now rejected with a dedicated error, per RFC 9758 Section 5.7. Previously such a registration was accepted and then silently shadowed forever by the admin route's precedence, a dead registration the application had no way to notice. The predicate lives in `node_ids` beside `get_admin_endpoint` and `is_local`, so the identity module owns every rendering of the concept.

Two smaller alignments: the unconfigured default's random allocator id is now tied by a named constant to the Experimental Use range (RFC 9758 Section 9.1), the range reserved for identities outside the coordinated allocation contract, and two unreachable error variants are removed. The `NodeIds` rustdoc now states the model: admin endpoints, at most one per scheme, everything else derived.

## Compatibility

Config files keep working via the alias. The one behaviour change is the new registration rejection, which converts a silent dead registration into a startup error. Pre-1.0; recorded in both CHANGELOGs.


ricktaylor (approved):
LGTM!!

PR #667, opened 2026-08-17
Refactor/grpc api v1

sylvain-pierrot:
# Redesign the gRPC API as a versioned v1 wire: per-method RPCs, a pure event stream, and a client-oriented proto crate

## Background

Hardy's BPA (Bundle Protocol Agent) is the daemon that stores, routes, and forwards DTN bundles (RFC 9171). Four kinds of external components run as their own processes and connect to it over gRPC: **applications** (send and receive payloads, called ADUs: Application Data Units), **endpoint services** (like applications, but exchanging whole raw bundles), **CLAs** (convergence-layer adapters, the drivers that move bundles over an actual link such as TCP), and **routing agents** (push routes into the BPA's routing table). The `hardy-proto` crate defines this gRPC protocol; this PR replaces that protocol and its implementation.

## What this changes and why

In the previous protocol, each component held exactly one long-lived bidirectional gRPC stream to the BPA, and *everything* travelled inside it: registration, every payload, every reply. Because a stream has no notion of request and response, the crate re-implemented that by hand: every message carried an id, both ends kept maps of pending replies, and background tasks matched answers to questions. That machinery duplicated what gRPC already does for ordinary calls (matching, deadlines, status codes, concurrency), and the single stream capped the protocol itself: a bundle had to fit inside one protobuf message (16 MiB), fully loaded in memory on both ends, queued behind everything else on the stream.

This PR replaces that wire entirely. **Each component type gets a real gRPC service: one `Subscribe` stream that carries only registration and small server events, while every action a component takes is an ordinary RPC, authorized by a session token, with payload bytes moving on their own chunked streaming calls.** The new schemas are versioned (`application.v1`, `endpoint.v1`, `cla.v1`, `routing.v1`) and use RFC 9171's own vocabulary throughout.

## The wire

A session works in three steps, and the schema is organized to make them visible:

1. The component opens `Subscribe` and sends `Register`; the BPA answers with a `Registration` event carrying its assigned endpoint id and a **session token**.
2. From then on the `Subscribe` stream only carries small events from the BPA: `Delivery` ("a bundle arrived for you: here is its id and size"), `BundleStatusReport` (RFC 9171's delivery/forwarding notifications), and for CLAs `Forwarding` ("please transmit this bundle"). Payload bytes never ride this stream.
3. Everything the component does is a separate RPC presenting the token: `Send` streams a payload up in 256 KiB chunks (metadata first), `Receive` collects an announced bundle as a chunk stream down, and the small calls (peers, routes, transfer outcomes) are plain unary RPCs.

Design decisions worth knowing before reading the schemas:

- **Delivery is announce-and-collect, not push.** The BPA holds an arrived bundle and announces it; the component fetches it when it chooses. A bundle never collected before the session ends is simply kept by the BPA and announced again to a later registration: exactly the delivery-deferral behaviour RFC 9171 specifies.
- **One identity everywhere**: the RFC 9171 bundle id keys announcements, collection, send results, and status reports. There are no invented transfer ids.
- **Cancellation exists exactly where gRPC has no native primitive, and nowhere else:**

| Who cancels what | Mechanism |
|---|---|
| Component aborts its upload | `cancel` message in the `Send` stream; the partial transfer is discarded |
| Component abandons a collection | `cancel` message in the `Receive` stream; the bundle stays held |
| BPA rejects mid-transfer | the RPC ends with a gRPC error status (native) |
| BPA withdraws a pending delivery | `cancelled` reply on the `Receive` stream |
| BPA deletes an accepted bundle later | `BundleStatusReport { DELETED }` (RFC 9171's own path) |

- **Errors are gRPC statuses, never message payloads.** A failed registration or rejected transfer terminates the call with a status, which is what every gRPC client library, proxy, and dashboard already understands; the vendored `google/rpc/status.proto` is gone.

## The limitations of the previous wire

Every one of these was structural: baked into the one-stream shape, not fixable by tuning.

- **A hard 16 MiB ceiling per bundle.** A bundle was one `bytes` field inside one stream message. Multi-gigabyte ADUs were unrepresentable on the wire at all.
- **Head-of-line blocking by design.** One first-in-first-out stream carried everything for a component: a single large or slow transfer stalled registration, deliveries, and status reports behind it.
- **Push-based delivery.** The BPA pushed the full payload whether or not the component was ready; the component had no say over concurrency, ordering, or memory.
- **A fragile shared fate.** One oversized or malformed message killed the whole stream, and a dead stream means the component is unregistered; the code needed defensive size checks just to protect the session from its own traffic.
- **No cancellation.** A transfer was one message: nothing could abort an upload in progress, decline a delivery, or withdraw a pending one.
- **No per-action semantics.** Deadlines, error codes, retries, and backpressure existed only for the stream as a whole, never per action.
- **A hand-rolled correlation engine.** Several hundred lines of message-id bookkeeping, pending-reply maps, and dispatcher tasks re-implementing what the transport already does, with their own deadlock and leaked-session failure modes (both found and fixed the hard way during earlier work).
- **No authentication.** Anyone who could reach the port could register as any component and speak for it; there was no credential to present and nowhere to hang an authorization decision.
- **Invisible to standard tooling.** Actions were variants inside an opaque stream message: no grpcurl, no per-method metrics or tracing, no interceptors.
- **No versioning.** Any wire change was a flag day for every deployed component.

## The performance impact

- **The payload ceiling is gone.** Transfers stream in 256 KiB chunks (with an 8 GiB reassembly guard), so multi-gigabyte ADUs are simply calls. Each transfer gets its own HTTP/2 stream and flow-control window, so a huge transfer no longer delays anything else.
- **Delivery is pull, not push.** The component collects on its own schedule and bounds its own memory; outgoing chunks are zero-copy slices of one held buffer.
- **Authorizing a call is one map lookup.** The session token itself is the key of a sharded concurrent map, so the per-RPC cost is a single hash probe: no per-call cryptography, no global lock. Possession of the token is the proof; a forged token is simply absent from the map.
- **gRPC does the work it was built for.** Matching, deadlines, per-call status codes, and concurrent handling come from the transport instead of a userspace re-implementation of HTTP/2 semantics inside HTTP/2.

## The optimizations this design unlocks

The wire shape was chosen so the following become possible later without touching the contract again:

- **End-to-end streaming into storage.** Announce-and-collect matches the planned pull-based evolution of the BPA's internal component API, so chunks can eventually flow from the wire into the bundle store without ever materializing in between: relaying a multi-gigabyte bundle then costs memory proportional to one chunk, not one bundle.
- **Component-scheduled transfers.** `Delivery` announces size and expiry up front, so a component can collect many bundles in parallel, prioritize small ones, or defer huge ones to a better contact window: scheduling policy lives in the component, where it belongs in a DTN.
- **Resume of interrupted transfers.** A collection keyed by bundle id extends to a ranged one with a single added field, a backward-compatible protobuf change; the old push model had no place for resume at all.
- **Reconnectable sessions.** The token is a signed JWT naming its registration, so the future evolution where a registration survives a dropped connection and the component reconnects with its token needs no rethink of the data plane.
- **A real authentication and authorization path.** The wire now has the two halves security needs: a place to authenticate (the `Register` handshake, plus TLS or mutual TLS on the connection, which the applications control since they own their gRPC server) and a credential to authorize with (every call presents the signed token). Per-method authorization, registration-time identity checks, and tenant separation all become configuration on top of this shape.
- **Per-method operational control.** Because every action is a real RPC, the standard gRPC toolbox applies per method: deadlines, retries, rate limiting on the data plane without touching the control plane, per-call compression, and per-method metrics and traces.
- **Independent evolution per surface.** Four versioned packages mean a future CLA-only change ships as `cla.v2` while applications keep speaking `application.v1`.

## The machinery impact

The correlation engine is deleted with nothing in its place, because the new wire makes it unnecessary by construction: after the handshake, nothing on the session stream expects a reply, so there are no pending maps, no message ids, and no dispatcher tasks on either side. The component side of a surface needs no engine at all, just a loop over incoming events; the BPA side keeps only real state: which sessions are live, which deliveries are held, and the key that signs tokens.

Ownership moves to match: **`hardy-proto` becomes the contract plus its two optional halves.** The default build is schemas, generated types, and the wire constants; the `server` feature adds the BPA-side bridges, which serve the four surfaces over any `hardy_bpa::bpa::BpaRegistration`, so any host embedding a BPA can offer the wire; the `client` feature adds the component SDK (`BpaClient`). `bpa-server` keeps only the transport composition: a validated `grpc.services` list chooses which bridges mount next to its health service. Every action is now a first-class RPC, visible to grpcurl, tracing, and interceptors.

## The Rust implementation

An overview of the direction the code takes; the same rule as the wire, keep only real state and let the type system carry the protocol:

- **A lock-free hot path.** Live sessions sit in a sharded concurrent map (`DashMap`) keyed by the session token, so authorizing a call is one hash probe and concurrent sessions never contend; set-once state lives in write-once cells, and no lock is held across an `await`.
- **Bytes move by reference.** Received chunks arrive as reference-counted buffers (`bytes::Bytes`) rather than copies, outgoing transfers are chunked as zero-copy slices of one buffer, and all waiting is event-driven (one-shot wake-ups, never polling).
- **One type per session, reading like the protocol.** Everything about a registration lives in a single struct, and its implementation of the BPA's component trait is a direct transliteration of the wire semantics: park and announce a delivery, await its collection, forward status reports, clean up on disconnect. Each RPC is one handler function, and a whole surface fits in one file.
- **The compiler checks the protocol.** Every wire `oneof` becomes a Rust enum, so message handling is an exhaustive `match` and a schema addition fails to compile every handler that ignores it; newtypes keep the token distinct from ordinary bytes; errors convert at exactly one point per call, gRPC statuses on the wire and the BPA's error types inside.

## Compatibility and state of the migration

The wire is a clean break with no interoperability with the old protocol, which is the point of the `v1` package versioning: a future incompatible wire ships as `v2` alongside a running `v1` instead of as another flag day. `hardy-proto` moves to 0.3 with a fully breaking API.

This PR lands the whole migration: all four v1 schemas; the server bridges behind `hardy-proto`'s `server` feature (registration with explicit and BPA-assigned service ids, chunked send with in-stream cancel, announce-and-collect delivery with abandonment, status report events, held-bundle semantics on disconnect, and the CLA `Forward` rendezvous with deferred transfer outcomes); the client SDK behind its `client` feature; and bpa-server composing the bridges behind its `grpc` feature. The application bridge is the template: the service, CLA, and routing bridges are copies of it with fewer or slightly different fields, so reviewing that surface deeply is reviewing them all. The remote-component binaries (`tools`, `tcpclv4-server`, `tvr`, and the workspace-excluded mtcp interop test) are migrated, and the in-process components (`bibe`, `echo-service`) adapt to the pull-based delivery trait.

This branch also lands the first step of the component API's pull-based evolution, because building the bridge showed the push API and the pull wire fighting each other: `on_receive` (push a payload, await the outcome) became `on_deliver` (announce a bundle id) plus `Sink::receive` (pull the bytes as a segment stream, where receiving the final segment is the completion and dropping the stream defers the delivery, RFC 9171's deferral for free). The wire did not change at all, which was the point of its shape: the bridge's held-delivery table and its invented delivery keys are simply deleted, `Delivery` events now carry real bundle ids, and every delivery flows announce, pull, complete end to end. The one declared scaffold left is the `Send` door's bounded accumulation; `SendMetadata.adu_size` already rides the wire as a preallocation hint, and the scaffold retires when the BPA can assemble application bundles without whole-payload buffering (a declared-authoritative size or streaming assembly).

## Deliberate asymmetries and named follow-ups

Stated here so review does not have to rediscover them:

- **`BpaClient` mirrors `BpaRegistration` but does not implement it.** Remote registration is fallible (connection, handshake) where the local trait is not, so the six register methods are inherent and local/remote polymorphism behind one trait is gone on purpose. Relatedly, `register_cla` carries no `EgressPolicy`: the v1 wire has no representation for egress policies yet, so remote CLAs run without one while the local trait keeps the parameter.
- **The component sinks' `cancel()` is not carried by the v1 wire** and returns an error saying so; in-stream cancel on the data-plane calls is the wire's cancellation.
- **BIBE's failure semantics change with pull-based delivery.** Collection completes the outer delivery, so a transient failure dispatching the decapsulated inner bundle now drops it where the push model kept the outer bundle parked for retry. The code owns this in a comment; revisit if BIBE needs stronger-than-best-effort semantics.
- **Security posture.** The session token is a non-expiring bearer credential and the gRPC listener has no TLS option yet, so deploy on loopback or a private network. The wire now has the two halves a real story needs (the `Register` handshake to authenticate in, the presented token to authorize with); TLS on the listener and registration-time identity checks are configuration to add, not wire changes.
- **Test state.** Every surface has an in-crate suite against a real BPA, including client-SDK roundtrips. The cross-crate lifecycle suite (BPA-initiated unregister, connection loss and keepalive-bounded death, simultaneous unregister, on every surface) is the named follow-up.


ricktaylor (approved):
LGTM!

PR #673, opened 2026-08-17
refactor(bpa): streaming dispatch seams for CLA ingress

ricktaylor:
Introduces the stream-shaped ingress seams for CLA-to-BPA dispatch: Segment (Next/Final) and Sink::dispatch_streamed, with ingress restructured to consume a segment stream and the accumulation kept behind the existing seam (dumb-concat interim — the deliberate placeholder until BundleStorage::save_stream lands).

Breaking: Sink gains a required method; every CLA implementation must provide dispatch_streamed.


bpa/src/dispatcher/mod.rs
```diff
@@ -12,6 +12,12 @@ mod reassemble;
 mod report;
 mod restart;
 
+// The default bound on a single reassembled bundle. Streaming producers
+// dissolved the transport-level caps that used to bound ingress implicitly,
+// so the concat chokepoint enforces one; sized generously above the old
+// 16 MiB wire cap to leave room for large ADUs.
+const DEFAULT_MAX_BUNDLE_SIZE: usize = 64 * 1024 * 1024;
```
sylvain-pierrot: You could use `NonZeroUsize` directly?

bpa/src/stream.rs
```diff
@@ -57,3 +59,234 @@ impl<T: Send + 'static> Sender<T> for hardy_async::channel::Sender<T> {
...
+pub async fn concat_stream<R: Receiver<Segment> + ?Sized>(
+    stream: &R,
+    max_size: usize,
+) -> core::result::Result<crate::Bytes, ConcatError> {
+    // The first segment is held as-is until a second arrives, so a
+    // single-`Final` stream (the whole-buffer convenience methods) is
+    // returned untouched — unconditionally zero-copy, even when the caller
+    // retains a clone of the `Bytes`.
+    let mut first: Option<crate::Bytes> = None;
+    let mut concat: Option<bytes::BytesMut> = None;
```
sylvain-pierrot: Small note: both types are from `bytes` ext crate, but one is a re-export. Maybe we shouldn't use the re-export inside the bpa itself?

bpa/src/stream.rs
```diff
@@ -104,6 +104,28 @@ impl<T: Send + 'static> Receiver<T> for hardy_async::channel::Receiver<T> {
...
+#[async_trait]
+impl<T: Send> Receiver<T> for CancellableReceiver<'_, T> {
+    async fn recv(&self) -> core::result::Result<T, RecvError> {
+        use futures::FutureExt;
+        futures::select_biased! {
+            _ = self.token.cancelled().fuse() => Err(RecvError),
+            r = self.inner.recv().fuse() => r,
+        }
+    }
+}
```
sylvain-pierrot, line 127: Could this be a token parameter instead of a decorator? `concat_stream(stream, max_size, cancel)` with the `select_biased!` inlined in the loop deletes `CancellableReceiver`; the dispatcher taking a bare `CancellationToken` doesn't leak registration knowledge, and cancellability becomes visible in the signature. Costs one parameter on `receive_bundle` (and the service door's equivalent) plus a never-fired token at the interim concat call sites.

ricktaylor: `concat_stream` is going to disappear over the next merges - so let's keep the decorator until all of them land

bpa/src/stream.rs
ricktaylor: I think I prefer using the re-export - I'll change

bpa/src/dispatcher/mod.rs
ricktaylor: Let me check if NonZeroUsize has const initialization, but yes, good idea

bpa/src/stream.rs
sylvain-pierrot: Alright! Let's keep that in the back of our minds

bpa/src/stream.rs
ricktaylor: Done

bpa/src/dispatcher/mod.rs
ricktaylor: Done

sylvain-pierrot (approved):
LGTM, great job!

PR #674, opened 2026-08-17
refactor(bpv7)!: structural parser + policy-free checks

ricktaylor:
The bpv7 API story: replaces the ParsedBundle/CheckedBundle/RewrittenBundle parse modes with a layered mechanism separating wire truth from policy — parse::parse (structural, keyless; BundleParser::push/finish for streaming callers) plus composable checks/rewrite primitives that produce facts and leave policy to the caller. The BPSec mechanism is unified along the way (OperationSet::check over BlockSet, one classify_unsupported, the single edit cascade), and the PICS conformance suite is ported against the new API with Bundle::semantic_eq.

Breaking: bpv7's parse surface is replaced wholesale; the rich bundle type moves to bpa. Callers compose their own pipelines from the primitives (the reference pipeline lives in bpa).

This is an interim change as part of a larger series


sylvain-pierrot (approved):
LGTM! Let's continue to land streaming 😁

PR #675, opened 2026-08-17
refactor(bpa,bpv7): pre-drain ingress gate and streaming-ready parse pipeline

ricktaylor:
The ingress rework on top of the structural parser: bundle parse splits into a pre-drain gate (header parse + keyed verify + lifetime/hop early-reject before any payload is spooled), block-1 BPSec verification defers for streaming, and oversized payloads get the Partial/PayloadTail signal. Conformance and hardening ride along: RFC 9172 §5.1.1 failure-drop applied on ingress instead of discarding the bundle, primary-block canonicalization for BPSec IPPT/AAD, reception reported before the duplicate check, RFC 9172 security reason codes in status reports, and the wire-derived-range bounds-check pass (32-bit safety).

Reviewer note: there is a known non-compiling window mid-branch (canonicality tuple-threading, restored by the canon-restore commit a few positions later) — the tip and CI are green; only bisection through the middle of this branch is affected.


sylvain-pierrot (approved):
LGTM! Amazing work

PR #676, opened 2026-08-17
perf(cbor,bpv7): allocation-free error paths and pre-sized rebuilds

ricktaylor:
Parse-path performance: non-bundle first bytes rejected without allocating, IncorrectType errors made allocation-free (semver-breaking for hardy-cbor, CHANGELOG'd), and the editor's rebuild vectors pre-sized. The structural-parser tests move to tests/parse.rs where they belong.


sylvain-pierrot (approved):
LGTM!!!

PR #677, opened 2026-08-17
refactor(bpa)!: structural bundle + metadata partitioned by write discipline

ricktaylor:
The metadata restructure: bundle.bundle becomes the structural bpv7 Bundle with decoded extension fields moving to metadata, and BundleMetadata is partitioned by write discipline — write-once provenance (received_at()/origin(), with Origin::{Ingress, Originated, Recovered} now persisted, fixing restart recovery mis-reading recovered bundles as locally originated), the WireCache group, a private Classification placeholder, and constructors (ingress()/originated()/new(); Default removed). Riding along: expired-at-ingress bundles drop silently with per-reason received.dropped labels, and reassembled ADUs carry the earliest fragment's provenance.

Breaking: the persisted metadata serde shape changes — stores written by earlier versions do not deserialize.


sylvain-pierrot (approved):
LGTM!!!

PR #680, opened 2026-08-17
 feat(bpa): streaming originate seam for the Service trait

ricktaylor:
Companion to #673: the service door gets the same stream-shaped seam as the CLA door. ServiceSink gains send_streamed (bundle delivered as stream::Segments), plain send funnels through the same streamed spine, and Segment moves to the shared stream module (re-exported from cla). A producer that drops its sender before Final cancels the send (StreamCancelled, nothing stored). The gRPC client sink concatenates and delegates to its unary send — the streamed RPC rides with the pending proto work this seam unblocks.

Breaking: ServiceSink gains a required method.


sylvain-pierrot (approved):
LGTM!

PR #682, opened 2026-08-19
feat(bpa): add Cla::forward_streamed, the streamed egress door

ricktaylor:
Adds the streamed egress primitive to the Cla trait — seam 6 of the six streaming seams (refactor_plan: "the pull-shaped Receiver/Segment stream is the target for all six seams"), and the first out-path seam to land. This is streaming_pipeline_design.md §6.2's Cla::write, landed as forward_streamed for consistency with dispatch_streamed/send_streamed, and keeping forward's bundle_id correlation parameter, which the sketch predates.

````rust
async fn forward_streamed(
    &self,
    queue: Option<u32>,
    cla_addr: &ClaAddress,
    bundle_id: &Id,
    stream: &dyn crate::stream::Receiver<Segment>,
    total_len: u64,
) -> Result<ForwardBundleResult>
````

Non-breaking: the method has a provided default body — the §6.2 buffered adapter — so all existing Cla implementors compile and behave identically. It collects via concat_stream capped at total_len, maps truncation → StreamCancelled and overrun → PayloadTooLarge, and delegates to forward(). Single-Final streams take concat_stream's zero-copy fast path, so buffered CLAs receive the identical Bytes.

32-bit safe: total_len is u64 with a usize::try_from pre-flight before any buffering — no as usize on a wire-scale length.
The BPA now always forwards through the streamed door (§6.2: "Internally, the BPA always uses write()"): the dispatcher wraps the loaded bundle as a single Final segment with total_len = data.len(), the exact mirror of Sink::dispatch's default body on the ingress side. All four ForwardBundleResult arms and the requeue fall-through are unchanged; a streamed-door Err takes the existing restore-to-Waiting path.

total_len contract: the exact byte count the stream will deliver — the framing hint length-prefixed transports need (e.g. TCPCLv4 XFER_SEGMENT), and the same number the v1 wire's Forwarding.bundle_size carries, so the internal seam and the future cla.v1 Forward bridge line up parameter-for-parameter.
Also: records the landed shape in a §6.2 "Landed state" note in streaming_pipeline_design.md (same convention as the §5 ingress note), and fixes a broken EgressPolicy intra-doc link in the trait docs (now zero rustdoc warnings in hardy-bpa).

# Why now
The public Cla trait is the extension surface — settling the streamed shape before more CLAs exist avoids a second breaking round. The default-body adapter means the streaming lands behind the API later (streamed storage load, egress executor) with no further trait churn.

# Tests
New bpa/tests/forward_streamed.rs (nothing added to pipeline.rs):

Full-path: a streaming CLA receives exactly one Final segment with exact total_len, bytes parse to the originated bundle; a forward-only CLA round-trips through the default adapter.

Full-path: a failed streamed attempt requeues via the existing restore/Waiting path and retries.

Direct-call: multi-segment reassembly; zero-copy single-Final (pointer-equality asserted); truncation → StreamCancelled with forward never invoked; overrun of total_len → PayloadTooLarge.

Gates: fmt --check, clippy --locked --all-targets --all-features -D warnings, test --locked --all-features --workspace all green (the two tcpclv4 socket-bind tests fail only in the sandbox used for development — AddrNotAvailable — and fail identically on clean main there).

# Merge notes (refactor train)
Deliberately shaped to stay out of the train's way: bpa/src/cla/mod.rs and stream.rs are untouched by refactor/parse/cbor-perf/metadata, tests live in a new file, and file-cla/tcpclv4/bibe/proto are untouched. Expected conflicts when the train lands: the single call-site hunk in dispatcher/forward.rs and the one doc blockquote — both mechanical.

# Follow-ups
Streaming overrides in tcpclv4 (XFER_SEGMENT framing from total_len) and file-cla, once the train lands.
BundleStorage::load_stream (the storage-side enabler; save_stream remains the in-path prerequisite).

The Deliver seam (announce/collect), re-ported from refactor/grpc-api-v1 onto the metadata model.
Eventually retire whole-buffer forward() per §5.1's "transitional convenience" note.

### DIFF
--- a/bpa/docs/streaming_pipeline_design.md
+++ b/bpa/docs/streaming_pipeline_design.md
@@ -785,6 +785,8 @@ Each step is a `mac.update()` or AAD accumulation call. The filter provides thes
 
 ### 6.2. CLA Egress: Cla::forward and Cla::write
 
+> **Landed state (2026-08, `feat/cla-forward-streamed`).** The streaming variant landed as `Cla::forward_streamed`, following the `dispatch_streamed`/`send_streamed` naming rather than the `write()` naming sketched below, and it keeps `forward`'s `bundle_id` correlation parameter, which the sketch predates. The non-streaming adapter is the trait's provided default body: it buffers the stream via `concat_stream`, capped at `total_len` (with a `usize` pre-flight for 32-bit targets), maps truncation to `StreamCancelled` and overrun to `PayloadTooLarge`, then delegates to `forward()`. The dispatcher now always forwards through the streamed door, wrapping the loaded bundle as a single `Final` segment with `total_len = data.len()`. The egress executor, streamed storage `load`, and the Transformer-derived `total_len` below remain pending.
+
 The existing `Cla::forward(Bytes)` method is retained for CLAs that expect a complete bundle in memory. A new streaming variant takes a `Receiver<Segment>` from which the CLA pulls chunks, mirroring `Sink::write` from §5.1:
 
 ```rust
--- a/bpa/src/cla/mod.rs
+++ b/bpa/src/cla/mod.rs
@@ -165,8 +165,8 @@ pub enum TransferOutcome {
 /// transport, such as TCP, UDP, or a custom link-layer protocol. It handles the
 /// transmission and reception of bundles over its specific medium.
 ///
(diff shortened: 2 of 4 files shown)


sylvain-pierrot (approved):
LGTM!

PR #683, opened 2026-08-19
feat(bpa): add Service::on_receive_streamed, the streamed delivery door

ricktaylor:
Adds the streamed delivery primitive to the Service trait — the low-level half of the delivery seams (seams 4/5 in refactor_plan), companion to #<CLA-PR> which landed seam 6 (Cla::forward_streamed). Same recipe: a default-bodied _streamed method so the change is non-breaking, and the BPA switched to always use the streamed door.

````rust
async fn on_receive_streamed(
    &self,
    stream: &dyn crate::stream::Receiver<Segment>,
    expiry: time::OffsetDateTime,
    total_len: u64,
) -> Result<()>
````

Non-breaking: the provided default body buffers via concat_stream capped at total_len (with a usize pre-flight for 32-bit targets) and delegates to on_receive(); single-Final streams take the zero-copy fast path, so existing services receive the identical Bytes.

Delivery semantics preserved: Err — including on a truncated stream — parks the bundle as WaitingForService, and re-registration on the same EID re-delivers, exactly as on_receive documents. A truncated delivery must never surface as Ok.
Dispatcher: the ServiceImpl::LowLevel arm of deliver_bundle wraps the loaded bundle as a single Final segment (total_len = data.len()), mirroring the ingress Sink::dispatch idiom.

Application delivery (Application::on_receive) is deliberately untouched — the payload door involves BPSec decode and is bound up with the announce/collect design, tracked separately.

# Tests
New bpa/tests/on_receive_streamed.rs: full-path single-Final delivery with exact total_len; buffered-service round-trip through the default adapter; failed streamed delivery parks and re-registration re-delivers; direct-call adapter tests (multi-segment reassembly, zero-copy pointer equality, truncation → StreamCancelled with on_receive never invoked, total_len overrun → PayloadTooLarge).

Gates green: fmt --check, clippy --locked --all-targets --all-features -D warnings, test --locked --all-features --workspace (the two tcpclv4 socket-bind tests fail only in the development sandbox, identically on clean main).

# Merge notes (refactor train)
Unlike the CLA branch, both source files here are touched by the train: the services/mod.rs trait hunk is a self-contained insertion (trivial resolution), but the local.rs call-site hunk sits inside deliver_bundle, which the metadata leg rewrites — expect to re-apply the one-arm change there when the train lands.

# Follow-ups
Streamed Application delivery, as part of the announce/collect Deliver design (from refactor/grpc-api-v1).
BundleStorage::load_stream so delivery can stream from storage instead of wrapping a loaded buffer.

Design-doc note for the delivery seams once the announce/collect direction is settled (§6-style "Landed state" note deliberately deferred — this is one half of seams 4/5).

### DIFF
--- a/bpa/src/dispatcher/local.rs
+++ b/bpa/src/dispatcher/local.rs
@@ -220,8 +220,16 @@ impl Dispatcher {
 
         let delivery_result = match &service.service {
             services::registry::ServiceImpl::LowLevel(svc) => {
-                // Pass raw bundle bytes to low-level services
-                svc.on_receive(data, bundle.expiry()).await
+                // Pass raw bundle bytes to low-level services, always
+                // through the streamed door: the whole bundle is in hand,
+                // so it travels as a single Final segment.
+                let total_len = data.len() as u64;
+                let (tx, rx) = hardy_async::channel::bounded(1);
+                crate::stream::Sender::send(&tx, crate::stream::Segment::Final(data))
+                    .await
+                    .trace_expect("bounded(1) send with live receiver cannot fail");
+                svc.on_receive_streamed(&rx, bundle.expiry(), total_len)
+                    .await
             }
             services::registry::ServiceImpl::Application(app) => {
                 // Extract and decrypt payload for Application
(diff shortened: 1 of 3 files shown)


sylvain-pierrot (approved):
LGTM!

PR #685, opened 2026-08-20
refactor(bpa): rename service delivery callbacks to on_deliver, passi…

ricktaylor:
…ng the bundle id

Rename Service::on_receive/on_receive_streamed to on_deliver/on_deliver_streamed and pass the delivered bundle's Id as the first argument. Rename Application::on_receive to on_deliver likewise, and add Application::on_deliver_streamed with the same buffered default adapter as the Service trait; the shared adapter body is factored into buffer_stream().

The dispatcher now delivers to Applications through the streamed door too, as a single Final segment, matching low-level Service delivery.

On the wire, AppReceiveRequest and ServiceReceiveRequest gain a bundle_id field using the same RFC 9171 key encoding as StatusNotifyRequest.


ricktaylor:
Push-with-result delivery makes completion atomic with the handler: the dispatcher marks a bundle delivered only after on_deliver returns Ok, and on the wire the client acks after its handler ran. The two failure windows inherent to announce-and-collect — a completion torn by consumer cancellation, and a bundle committed that the client never saw — cannot occur in this model, by construction.

sylvain-pierrot (approved):
LGTM!!!

PR #692, opened 2026-08-24
fix(bpa): claim terminal bundle resolutions with a conditional tombstone

ricktaylor:
The expiry reaper resolves bundles regardless of which task currently owns them, and both it and the owners resolved unconditionally: a bundle expiring while its delivery (or any drop) was in flight could be resolved twice, emitting contradictory deletion reports to report_to ("Lifetime expired" from the reaper, then delivered + "No additional information" from the completing delivery) and double-decrementing the status gauge.

Every terminal resolution is now a compare-and-swap claim, with the losing side staying silent:

- drop_bundle claims via Store::tombstone_if before reporting, covering every drop site (reaper expiry, filter drops, admin, depleted storage).
- deliver_bundle's completion claims the same way, kept apart from drop_bundle so a delivery never counts as a dropped bundle.
- deliver_bundle's Err park swaps conditionally out of Dispatching, so a deferred delivery cannot resurrect a bundle the reaper has tombstoned.

This mirrors the conditional-tombstone shape the CLA transfer-outcome path already uses, independently arrived at by refactor/grpc-api-v1's complete_delivery. The race matters most once deliveries can be held open: an announce-then-pull bridge holds toward expiry, where the two resolvers practically synchronise.

The regression test parks a bundle (arming the reaper's watch list), redelivers into a service that holds the stream across expiry, and asserts report_to sees exactly one deletion report.

### DIFF
--- a/bpa/src/dispatcher/forward.rs
+++ b/bpa/src/dispatcher/forward.rs
@@ -46,10 +46,16 @@ impl Dispatcher {
         let (pre_rewrite, data) = match self.update_extension_blocks(&bundle, data) {
             Err(e) => {
                 warn!("Failed to update extension blocks: {e}");
-                self.store
-                    .update_status(&mut bundle, &bundle::BundleStatus::Waiting)
-                    .await;
-                return self.store.watch_bundle(bundle).await;
+                // Conditional: the reaper can resolve the claimed bundle at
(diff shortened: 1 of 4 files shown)


sylvain-pierrot (approved):
LGTM!

PR #693, opened 2026-08-24
ci: parallelize the jobs and cache protoc

sylvain-pierrot:
# Restructure the CI into parallel jobs with a cached protoc

## Background

The Rust workflow ran six jobs: `build` (format, lint, compile the tests), then `test`, `storage-tests`, and `ping-tests` fanned out behind it via `needs`, with `no-std` and `msrv` on the side. The fan-out shape meant every downstream job paid the full runner preamble (checkout, toolchain, ~1 minute of `rust-cache` restore, and an apt protoc install) to run a few seconds of already-compiled tests, and nothing test-related could start until `build` finished. Five of the six jobs installed protoc with `apt-get update && apt-get install`, where the `update` alone costs 20 to 40 seconds per job.

## What this changes and why

The six jobs become four fully independent ones, with no `needs` chains: wall time is the slowest job, not a build-then-fan-out sum.

| Job | Runs | Cache key | Protoc |
|---|---|---|---|
| `checks` | `cargo fmt --check`, `cargo clippy -D warnings` | `checks` | yes |
| `tests` | the whole workspace test run (storage suites included), then the ping e2e | `tests` | yes |
| `no-std` | 32-bit bare-metal compile checks | `no-std` | no |
| `msrv` | `cargo check` on the MSRV toolchain | `msrv` | yes |

Two decisions carry the design:

- **Lint and test split into parallel jobs because they share no build artifacts anyway.** Clippy produces check-profile metadata; `cargo test` produces debug objects. Running them on two runners duplicates almost no compilation, unlike the old fan-out, which restored the same debug artifacts onto three runners. Each job saves its own `rust-cache` key, so neither clobbers the other's cache.
- **Everything that shares the debug target dir stays in one job.** The workspace tests, the service-backed storage suites, and the ping e2e all consume the same build, so they are sequential steps of `tests` rather than jobs: splitting them is what made the old layout pay three runner preambles for seconds of work. `--all-features` activates `storage-tests`' `postgres,s3` features, so one `cargo test --workspace` invocation now covers what used to be two jobs, with postgres and minio brought up beside it.

## The protoc install

The apt install is replaced by one `actions/cache` entry keyed on a `PROTOC_VERSION` workflow variable and shared by every job that needs protoc. On a cache hit, which is every run except the one after a version bump, the binary materializes with no network access at all. On a miss, the fill is a direct download of the GitHub release asset: that URL is plain CDN, not the GitHub API, so it is not subject to the API rate limits that actions like `arduino/setup-protoc` hit when they list releases. Bumping `PROTOC_VERSION` is the whole upgrade procedure.

## The cost and wall-time impact

- Three runner preambles per PR run are gone (checkout, toolchain, cache restore, protoc, times the three old fan-out jobs), and with them four `apt-get update` runs.
- Nothing waits on a `build` job anymore: `checks` and `tests` start together, and a formatting failure still surfaces in the first minute because it is the first step of `checks`.
- One cache save per key instead of one save plus three restores of the same key.

## Deliberate choices and named follow-ups

- **Branch protection must be updated with the merge:** the required checks are now `checks` and `tests`; the old `build`, `test`, and `storage-tests` check names no longer exist.
- **`no-std` and `msrv` stay separate** because they use a different target and a different toolchain respectively: merging either into the main jobs would poison the shared caches.
- **`cargo-nextest` is the next wall-time lever, not taken here.** It would parallelize the workspace's many test binaries inside `tests` (with `cargo test --doc` kept alongside for doctests), but it changes the test runner's execution model, so it deserves its own change.
- The docker services start with `up -d --wait` immediately before the test step; if image pulls ever dominate, the pull can overlap the earlier steps, a readability trade that is not worth it today.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,9 +19,10 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
+  PROTOC_VERSION: "29.3"
 
 jobs:
-  build:
+  checks:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


ricktaylor (approved):
LGTM

PR #694, opened 2026-08-24
fix(bibe)!: encode the encapsulated bundle as a byte string

sylvain-pierrot:
# Fix the BIBE PDU wire encoding: byte-string payloads and honest offsets

## Background

BIBE (Bundle-in-Bundle Encapsulation) tunnels a whole bundle as the payload of another bundle: `encapsulate` wraps an outgoing bundle into a BIBE PDU, and the decapsulation service unwraps arriving PDUs and re-dispatches the inner bundle. The PDU is a CBOR array of a transmission id, a total length, a segmented offset, and the encapsulated bundle bytes.

## What this changes and why

Writing the first encapsulate/decapsulate round-trip test exposed that the wire format never worked end to end:

- **The inner bundle was encoded as a CBOR array of integers, one uint per byte, not as a byte string.** `a.emit(inner.as_ref())` picks hardy-cbor's slice-of-integers encoding; the byte-string encoding needs the `encode::Bytes` wrapper. Decapsulation rejects arrays, so every tunneled bundle died at the far end.
- **Decapsulation applied the byte-string range to the whole payload instead of the item.** The range returned by `parse_value` is relative to the item start, so the recovered "bundle" was shifted by the PDU header length: it included header bytes at the front and lost the tail.

Two hardy nodes could exchange BIBE bundles only because both were equally wrong; a spec-correct peer could not interoperate with hardy at all. **The fix is breaking against the previous hardy-bibe wire format**, which is unavoidable: the old format was the bug.

Additionally, `decapsulate` treated every PDU as complete, silently accepting fragments: a non-zero transmission id, total length, or segmented offset now fails with the new `Error::SegmentedPdu`, because segmentation is declared future work in the design doc and treating a fragment as a whole bundle corrupts the payload. The existing permanent-failure path in `on_deliver` handles the rejection.

## Tests

Byte-identity round-trips across inner payload sizes chosen to cross every CBOR length-field boundary (0, 1, 23, 24, 255, 256, 65535, 65536), outer bundle field assertions, and malformed-PDU cases: each non-zero header field, trailing garbage, a non-byte-string segment, and a wrong-arity array. `cargo test -p hardy-bibe`, clippy with `-D warnings`, and fmt are clean; the design doc is updated to state the complete-PDU-only contract.

### DIFF
--- a/bibe/src/cla.rs
+++ b/bibe/src/cla.rs
@@ -61,7 +61,7 @@ impl BibeCla {
     }
 
     /// Encapsulate an inner bundle into an outer bundle.
-    fn encapsulate(&self, inner: Bytes, outer_dest: Eid) -> Result<Bytes, Error> {
+    pub fn encapsulate(&self, inner: Bytes, outer_dest: Eid) -> Result<Bytes, Error> {
         // Parse inner bundle to get lifetime for outer bundle
         let parsed = ParsedBundle::parse(&inner, bpsec::no_keys)?;
         let lifetime = parsed.bundle.lifetime;
@@ -73,7 +73,11 @@ impl BibeCla {
             a.emit(&0u64); // transmission-id
             a.emit(&0u64); // total-length
             a.emit(&0u64); // segmented-offset
-            a.emit(inner.as_ref()); // encapsulated-bundle-segment
+
+            // encapsulated-bundle-segment, as a definite-length byte string
+            // (a bare `&[u8]` would encode as a CBOR array of integers, which
+            // decapsulation rejects)
+            a.emit(&hardy_cbor::encode::Bytes(inner.as_ref()));
         });
 
         let (_bundle, data) =
--- a/bibe/src/cla.rs
+++ b/bibe/src/cla.rs
@@ -1,23 +1,21 @@
-use alloc::borrow::Cow;
+use alloc::boxed::Box;
 use core::num::NonZeroU32;
 
 use hardy_async::sync::spin::Once;
 use hardy_bpa::{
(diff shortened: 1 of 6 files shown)


bibe/src/service.rs
```diff
@@ -27,7 +27,7 @@ impl DecapService {
     }
 
     /// Extract inner bundle from outer bundle payload.
-    fn decapsulate(&self, outer_bytes: Bytes) -> Result<Bytes, Error> {
+    pub fn decapsulate(&self, outer_bytes: Bytes) -> Result<Bytes, Error> {
```
ricktaylor: Why pub?

bibe/src/cla.rs
```diff
@@ -61,7 +61,7 @@ impl BibeCla {
     }
 
     /// Encapsulate an inner bundle into an outer bundle.
-    fn encapsulate(&self, inner: Bytes, outer_dest: Eid) -> Result<Bytes, Error> {
+    pub fn encapsulate(&self, inner: Bytes, outer_dest: Eid) -> Result<Bytes, Error> {
```
ricktaylor: Why pub?

ricktaylor (commented):
Small questions...

bibe/src/cla.rs
sylvain-pierrot: Because of a test leaving into `service.rs` :/
I’ve extracted `encapsulate()` and decapsulate()`  it into a dedicated file `pdu.rs`

ricktaylor (approved):
LGTM

PR #695, opened 2026-08-24
fix(cbor): enforce the PrecisionLoss contract when narrowing floats

sylvain-pierrot:
# Enforce the PrecisionLoss contract and close the codec's untested edges

## Background

hardy-cbor is the workspace's RFC 8949 canonical CBOR codec: every bundle byte hardy parses or emits goes through it, so a silent decoding defect here is a wire defect everywhere. A mutation-style audit of its tests (for each assertion: name a plausible code mutation and check whether any test fails) found one production bug and a set of edges no test pinned.

## What this changes and why

**The production bug: `Error::PrecisionLoss` could never fire.** The float-narrowing decoders relied on `FromPrimitive::from_f64` returning `None` on lossy conversions, but num-traits and half implement it as an `as`-style cast that always returns `Some`: decoding the f64 encoding of 1.1 as f32 silently returned a rounded value, violating the documented contract. Narrowing now does a widening round-trip equality check (the same pattern the header parser already used), with an explicit NaN passthrough since NaN never compares equal to itself yet is representable at every width.

**The closed test gaps, each verified to kill a named mutation:**

- `parse_exact`, the documented guard against trailing bytes smuggled past a block decoder, had zero tests; the mutation `len == data.len()` to `<=` survived the whole workspace. Now pinned for scalars and containers.
- Non-canonical float detection was only tested via Inf/NaN, which a separate match arm handles; emptying the Normal/Subnormal downgrade arms survived every test. Vectors like 1.5-as-FP32 now pin them.
- The definite-length sequence over/underfill panics, `Series::try_new` count overflow (`TooBig`), chunked-string rules (tagged chunks are `InvalidChunk`, non-minimal chunk lengths clear `shortest`), `Option<T>` decoding, tag-folding into `shortest`, and map `count()` pair division all gained their first tests.

## Test hygiene

The RFC 8949 mega-tests are split into focused per-concern tests so a failure isolates; the `*emit(&x).0` pattern repeated about 60 times is behind a named helper; weak `contains` assertions on debug output became full-string equality; and the header fast-path regression tests moved from the integration suite into an inline test module, since they target private functions.

## Verification

All suites green, clippy `-D warnings` and fmt clean. The key fixes were spot-checked by mutation: the bug or mutation was applied, the new test observed failing, and the change reverted.

### DIFF
--- a/cbor/src/decode/head.rs
+++ b/cbor/src/decode/head.rs
@@ -554,3 +554,95 @@ fn parse_uint_minor_fast(minor: u8, data: [u8; 8]) -> Result<(u64, bool, usize),
         _ => Err(Error::InvalidMinorValue(minor)),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use hex_literal::hex;
+
+    use super::*;
+
+    // parse_uint_minor_fast must agree with parse_uint_minor on value,
+    // shortest flag, and byte count for every minor form.
+    #[test]
+    fn uint_minor_fast_matches_slow() {
+        for (minor, bytes) in [
+            (0u8, hex!("0000000000000000")),
+            (23, hex!("0000000000000000")),
+            (24, hex!("7B00000000000000")), // 123, canonical
+            (24, hex!("1700000000000000")), // 23, non-canonical
+            (25, hex!("0100000000000000")), // 256, canonical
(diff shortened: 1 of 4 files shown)


ricktaylor (approved):
LGTM!

PR #696, opened 2026-08-24
fix(eid-patterns): canonicalize pattern-to-EID conversion

sylvain-pierrot:
# Canonicalize pattern-to-EID conversion and pin the matching semantics

## Background

hardy-eid-patterns implements the EID pattern language the RIB matches routes against. Its conversion and comparison semantics sit under every routing decision, and an audit found three behaviors where the crate disagreed with itself or with hardy-bpv7, plus core match arms no test had ever exercised.

## What this changes and why

Three production fixes, each judged unambiguous against the crate's own documentation and hardy-bpv7's parser:

- **`IpnPatternItem::try_to_eid` now canonicalizes exactly like the bpv7 EID parser**: `(0,0,0)` is `Eid::Null`, `(0, u32::MAX, s)` is `Eid::LocalNode`, and `(0,0,s != 0)` is rejected as bpv7 does. Previously it produced non-canonical `Eid::Ipn` values that could never equal a parsed EID, and the null-endpoint cross-scheme subset arms were dead code; `is_subset("ipn:0.0.0", "dtn:none")` now holds both ways.
- **A range that collapses to one value parses as `Single`**, fixing the inconsistency where `ipn:1.[5]` and `ipn:1.5` displayed identically but compared unequal by `Eq`, `Hash`, and serde.
- **`TryFrom<EidPattern> for Eid` accepts a set whose items all name the same EID**, which the two-spelling null set produced by `From<Eid::Null>` requires.

The new tests pin what was previously unguarded: `EidPattern::Any` matching every scheme, the `Ord` structural tiebreaker for equal-score patterns (without it, a `BTreeSet` of routes silently drops one), union specificity with unequal member scores, the `Eid::Null` match arm, interval scoring boundaries, `expand_local_node` on mixed sets, a full-corpus Display/FromStr round-trip, `is_subset` including the `(None, Any)` arm, and numeric boundary parsing (u32 overflow, zero and overflowing scheme numbers).

## Test placement

`src/tests.rs` is dissolved per the repo's placement rule: structural parse-shape tests moved inline into the modules they test (`ipn_pattern.rs`, `dtn_pattern.rs`, `parse.rs`), and all behavior tests through the public API moved to `tests/patterns.rs`. Every original assertion is preserved.

## Named follow-ups

- dtn demux matching is currently case-insensitive; RFC 9171 defines the demux as case-sensitive. Deliberately left unpinned rather than tested as correct: it needs a maintainer decision.
- The fuzz target's Display round-trip check is commented out as "too strict for the current parser"; with the single-value-range normalization it may now be enableable.


bpa/src/routing/table.rs
```diff
@@ -297,18 +297,9 @@ fn sorted_insert<'a>(peers: &mut Vec<(u32, &'a Eid)>, peer: u32, next_hop: &'a E
 
 // Route selection iterates patterns in specificity order and must compare the
 // specificity of the pattern that *matched*, so a multi-item union is never
-// stored as a single key: any set-level score ranks every member by one
-// aggregate, letting a broad member drag a specific sibling behind routes the
-// sibling strictly beats. A union route is shorthand for one route per member.
+// stored as a single key; `EidPattern::into_atoms` owns that decomposition.
 fn flatten(pattern: EidPattern) -> Vec<EidPattern> {
```
ricktaylor: Can we not just inline this function in the call site?

ricktaylor (changes requested):
LGTM!!

bpa/src/routing/table.rs
sylvain-pierrot: absolutely!

ricktaylor (approved):
LGTM!!

PR #697, opened 2026-08-24
Cover bpv7's wire-input structural rules and harden BPSec key hygiene

sylvain-pierrot:
# Cover bpv7's wire-input structural rules and harden BPSec key hygiene

## Background

hardy-bpv7 parses untrusted wire input, so its tests are the workspace's main defense against malformed-bundle bugs. An adversarially verified audit found several structural rules RFC 9171 imposes on incoming bundles were untested — the code enforced them, but no test would fail if any of them regressed. Closing those gaps, the same review surfaced key-handling defects in the BPSec production code, fixed here rather than left for a second pass. A follow-up adversarially verified review of this branch then produced a second tranche: a typed error chain, several make-invalid-states-unrepresentable refactors, and a round of test and comment fixes, all included below. Every asserted wire behavior matched RFC 9171/9172.

## Wire-input coverage

**Structural block rules through `parse::parse`**, driven by splicing helpers in `tests/common/mod.rs` that build malformed bundles byte-for-byte (each layout commented): duplicate payload blocks, duplicate block types that must be unique, duplicate block numbers, payload not final, and missing payload, each asserting the typed error (`DuplicateBlockNumber`, `DuplicateBlocks`, `MissingPayload`, and `NotCanonical` where the canonical-form check fires first). The status-report reason code these violations map to is the BPA's concern, not the parser's, so it is not asserted here.

**CRC framing and BPSec target rules** (split into `crc_rules` and `bpsec_rules` modules): a block declaring a CRC type without a CRC value and a spurious CRC value with type none (both pinned to the exact `InvalidField { field: "block" }` over `InvalidCBOR(AdditionalItems)` chain), an unrecognised CRC type (`InvalidCrc(InvalidType)`), a BCB targeting the primary block (`InvalidBCBTarget`), and a BCB targeting the payload without must_replicate (`BCBMustReplicate`).

**Previously dead parsing paths**: fragment primary blocks (offset boundaries kill the off-by-one mutants; `InvalidFragmentInfo` both direct and through the full parse, plus the non-fragment-with-fragment-fields arm, pinned to its exact CRC-value error chain), the BPSec abstract-syntax-block structural errors (no targets, duplicate targets, mismatched result counts in both directions, invalid security sources — unit tests beside the internal `AbstractSyntaxBlock`), BCB-AES-GCM IV-length bounds with full 8–16 roundtrip coverage and both out-of-range boundaries pinned at the parse seam, and status-report assertions with real timestamps including the `[true, 0]` zero-time wire form.

**False positives repaired**: the status-report suite could not detect a received/forwarded field swap (a positional round-trip loop over the four assertion slots now can); `invalid_crc` corrupted an arbitrary byte and accepted any error (it now corrupts the stored CRC value for both CRC types and asserts the `IncorrectCrc` chain); the version-rejection assertion's byte-pattern scan is hard-anchored so it fails loudly if the fixture drifts; the shared-BCB failure-drop test keeps the pinned RFC 9173 Appendix A.4 keys and first proves the pristine vector verifies cleanly, so its deliberate ciphertext corruption is demonstrably load-bearing; the tagged-dtn-SSP test now pins `NotCanonical` through the typed error chain instead of a field name its structural-error twin also matches.

## BPSec key hygiene (production)

- **Raw symmetric key bytes live in one type.** `bpsec::key::SecretBytes` holds the JWK `k` bytes: zeroized on drop, length-only `Debug` (so `Type`, `Key`, and `Error::InvalidKey` can simply derive), no `Display` at all, base64url serde implemented on the type itself, and a single greppable `expose_secret()` seam for cryptographic use. Construction and deserialization never let plaintext transit an unguarded buffer: input moves into a zeroize guard before any conversion, and the JWK path decodes via `decode_slice` into a pre-zeroed buffer. `Type::octet_sequence` builds keys from raw bytes without exposing the wrapper (consumer fixtures in bpa and bpa-server converted).
- **MAC verification is constant-time by construction.** The BIB result is a `MacTag` whose bytes are private to its module: verification goes through `Mac::verify_slice`, and the only `PartialEq` a caller can reach is `subtle::ConstantTimeEq`, so a short-circuiting `==` on tag bytes no longer compiles (this also makes the bundle-equality and tools comparison sites timing-safe).
- **The unwrapped CEK never exists outside a zeroize guard.** `key_wrap::unwrap` allocates its output buffer inside `Zeroizing` while still all-zero, so every exit path (including `?`) wipes in `Drop`; the previous exact-size ritual, `debug_assert`, and manual error-path wipe are deleted because the ownership makes them redundant.

## Enforce-by-construction refactors

- **Typed field-error chain.** Every domain's `InvalidField` now boxes its own error type (`Box<Error>`, not `Box<dyn Error>`), with foreign leaf errors converted at wrap time through `From` impls. The streaming parser's feed-me-more extraction becomes a total match (a newly wrapped domain is a compile error, not a silently missed `NeedMoreData`), `require_canonical` loses its downcast sniff, and every test asserts nested error patterns instead of walking `downcast_ref` ladders.
- **`HopInfo.limit` is `core::num::NonZeroU8`.** The RFC 9171 §4.4.3 range 1..=255 is structural on every producer path: builder, JSON `BundleTemplate`, `bp create --hop-limit`, and `bp ping --ttl` (which previously emitted RFC-invalid hop limits unchecked) all reject out-of-range values at the type/argument layer, and the wire parser keeps the one boundary conversion.
- **`Eid::Unknown` carries a validated `UnknownSsp`.** The stashed scheme-specific part is guaranteed to be exactly one well-formed CBOR item, so re-emit is provably safe and `Display` cannot fail; the garbage-fallback arm, its `DisplayError` enum, and the test that pinned the incoherent state are deleted.
- **The BCB IV is an 8–16 byte enum.** `Parameters.iv: Iv` makes an out-of-range IV unrepresentable; `Iv::from_bytes` is the sole wire seam where `InvalidIvLength` can arise, and `decrypt_gcm` dispatches exhaustively, handing each exact-size array to its `AesGcm<Aes, Un>` nonce infallibly.
- **One ASB unpacking path.** `AbstractSyntaxBlock`'s range fields are private; `into_operations` is the only consumer, so every security context gets parse-parameters-once, Arc-share, per-target results keyed by wire target number, and `InvalidField` wrapping without re-implementing the loop (the three previous verbatim copies collapse into declarative calls).
- **No duplicated key-algorithm dispatch in BCB encrypt.** The wrap primitive is bound once as a function pointer in the same match that validates the algorithm, deleting the second `match jwk.key_algorithm` and its `unreachable!` arm; the variant dispatches keep `Unrecognised` arms that return errors rather than panicking.

## Test-key policy

No test hard-codes cryptographic values. Every arbitrary literal JWK key in the signer, checks, editor, rfc9173, and bpa/bpa-server suites is replaced with a freshly generated value of the same length, via a shared `rand_k` helper in `tests/common` backed by the same OS RNG production uses. A key stays verbatim only where its value is part of the fixture's contract — the RFC 9173 appendix vectors, whose tests verify spec-derived bytes (including the Appendix A.4 pair in the shared-BCB failure-drop test). Key agreement is explicit data flow (each test binds the generated key once and passes it to both the signing and verification paths), and every wrong-key test generates both keys and `assert_ne!`s them.

These rules — the pinned-vector criterion for hard-coding, generated fixtures, constant-time verification, `SysRng`-only cryptographic randomness, key bytes born inside zeroize guards, secrets never in `Display`/`Debug`/logs — are recorded as hard requirements in the code and test style guides and the AGENTS.md summary.

## Test hygiene

The `hop_info` and `status_report` inline test modules move out of `src/` into `tests/` per the placement rule (they exercise only public API). `tests/eid.rs` is rewritten from three monolithic table tests into focused named tests, with the `str_ipn` checks made stricter (`Eid::Ipn` exactly, not `Ipn | LegacyIpn`). Fixtures shared between integration-test binaries live in the cargo-conventional `tests/common/mod.rs`, and the primary-block splice logic exists in exactly one place (`end_of_primary` / `insert_after_primary`), consumed by every suite that byte-splices. The `eid_cbor` fuzz target gains a re-emit/re-parse oracle plus a byte-identity check for canonical unknown-scheme EIDs, and the fuzz post-mortem tests fail loudly on unreadable artifacts instead of passing silently. `bundle_tools_test.sh` builds the CLI tools once (target dir via `cargo metadata`) and runs its 26 cases against the built binaries in a private `mktemp` directory with an EXIT-trap cleanup, instead of ~50 `cargo run` invocations in a shared fixed directory.

## API compatibility

Breaking changes for out-of-tree consumers, each with a mechanical migration: `Type::OctetSequence`'s field is `SecretBytes` (construct via `Type::octet_sequence` / `SecretBytes::new`; read via `expose_secret()`); `HopInfo.limit` is `NonZeroU8` (and `HopInfo` is now `Copy`); `Eid::Unknown`'s payload field is `ssp: UnknownSsp` (construct via `UnknownSsp::new`, read via `as_cbor()`); `bib_hmac_sha2::Results` wraps a `MacTag`; `bcb_aes_gcm::Parameters.iv` is the `Iv` enum; `AbstractSyntaxBlock`'s parameter/result ranges are private behind `into_operations`; and code matching through `InvalidField` sources pattern-matches typed boxes instead of downcasting. One new production dependency: `subtle` (no_std, behind the `rfc9173` feature); bpa-server gains a `rand` dev-dependency for generated test fixtures.

## Verification

All bpv7 suites green (196 tests across the lib, the integration binaries, and doctests). Workspace-wide `cargo fmt --check`, clippy with `-D warnings`, and `cargo test --locked --all-features --workspace` pass (the postgres storage suites require a live PostgreSQL and are environment-gated). The fuzz crates compile and the tools script passes end to end.


bpa-server/src/bpsec.rs
```diff
@@ -331,9 +331,7 @@ mod tests {
     fn hmac_key(kid: &str) -> Key {
         Key {
             id: Some(kid.into()),
-            key_type: Type::OctetSequence {
```
ricktaylor, line 334: Please don't rename OctetSequence - it's what it's called in the JWK standard and I don't want to deviate

ricktaylor (changes requested):
Looks pretty good - I have some incoming changes for hmac_sha2 - but I'll rebase over this

bpa-server/src/bpsec.rs
sylvain-pierrot: The variant is still there :)

ricktaylor (approved):
LGTM!!

PR #698, opened 2026-08-24
fix(bpa): re-open the storage channel fast path while the consumer drains

sylvain-pierrot:
# Fix the storage channel's re-open liveness and make the BPA test suite event-driven

## Background

This started from a real CI failure: `storage::channel::tests::test_hysteresis_recovery` timed out waiting for the channel to re-open. The diagnosis found a production liveness gap, and the wider sweep it triggered found the BPA's tests relying on sleeps for synchronization, several tests that could not fail when the behavior they name broke, and untested core paths in routing, filtering, and recovery.

## The production fix

The hybrid storage channel closes its fast path when the in-memory buffer fills and re-opens it once the buffer drains to half capacity. The poller checked that condition exactly once, right after its storage drain, then parked on a notification that consumer progress never raises: if the drain finished while the buffer was still above threshold, the fast path stayed closed until the next send. The poller now re-checks on a short tick while it sits in that drained-but-full state; the tick costs nothing in the steady state. Two `cfg(test)` counters (congestion CAS wins, completed drain cycles) make the state machine assertable without touching production behavior.

## No timing-based synchronization

Every sleep-as-synchronization in the BPA tests is replaced by an event: a `ParkSignallingStore` metadata decorator raises a channel event at the exact moment the dispatcher parks a bundle as `WaitingForService` (previously a 100 ms sleep guessed at it); the parked-send tests prove the consumer is inside the stream by a capacity-1 channel rendezvous; restart-recovery tests wait on status-write and deletion signals from decorated stores. Timeouts remain only as hang failsafes, commented as such. The throughput test's wall-clock gate (>1000 bundles/sec, the original full-suite flake) is opt-in via `HARDY_TEST_PERF_GATE` for a dedicated perf job; REQ-13 stays formally verified by the criterion benchmark.

## Tests that could not fail, now real

The reaper cache tests re-implemented `Reaper::watch`'s algorithm on a local `BTreeSet` and never called production code; they now drive the real reaper through a snapshot accessor and assert the wakeup with `poll!`. `test_lazy_expiry` proves the slow-path expiry filter (it previously admitted in a comment that it proved nothing); the congestion test pins the actual CAS; `impacted_vias` is finally called by the test named after it, with both inversion mutants killed.

## New coverage

First tests for the auto-registered ingress validity filters (lifetime, hop limit, CRC/BIB presence, Bundle Age), filter-chain write paths and cross-level ordering, ECMP hash spread (including making the flow label live in the hash), the corrupt/duplicate/orphan/Dispatching restart-recovery branches, `max_bundle_size` enforcement at the CLA ingress door, peer-id allocation across wraparound, and the RIB remove/reset side effects. The fuzz harness gets a per-process temp dir so a crashed run cannot replay stale state into fuzz reproductions.

## Test hygiene

Shared scaffolding is consolidated into `bpa/tests/common/` (bundle builders, node-id helpers, event receivers, the signalling stores); public-API tests moved out of `src/` into new integration files per the placement rule; the fuzz post-mortem tests inlined into the harness's lib.

## Named follow-up

The storage channel spawns a task on its pool during shutdown, discovered when enforcing the task pool's documented spawn-after-shutdown panic aborted a channel test. The async crate's docs now state actual behavior; the call site here deserves its own look.


ricktaylor:
I'm not going to merge this as the internal storage channel and queuing is under major rewrite.  However - let's keep this open to ensure we match the fixes

PR #699, opened 2026-08-24
fix(bpa-server): construct storage config defaults per field

sylvain-pierrot:
# Fix feature-reduced storage config and make the bpa-server e2e scripts assert real outcomes

## Background

bpa-server's e2e shell scripts existed but proved nothing: `test_bundle_processing.sh` nested its file-cla settings under a key the CLA never reads and exited 0 unconditionally, and `test_static_routes.sh` asserted only that the process was still alive after each reload. Reviving the first script uncovered a production config bug.

## The production fix

`StorageConfig` used a container-level `#[serde(default)]`, which constructs the complete `Default` up front, and the backend selector defaults deliberately panic in builds without the default-backend features. The result: a build without `sqlite-storage` panicked at startup even when the config explicitly selected `storage.metadata: memory`, contradicting the comment at the panic site. Defaults are now per-field, so they fire only for keys actually absent: explicitly configured backends load in any build, and the unconfigured default still refuses to silently degrade to the memory backend, which the existing `should_panic`-gated test continues to pin.

## The scripts

- `test_bundle_processing.sh` builds with `--no-default-features --features file-cla` (its original intent, viable again with the fix), nests the file-cla keys where `hardy_file_cla::Config` reads them, polls outbox consumption and peer-inbox delivery with deadlines, and exits nonzero on failure. Verified end to end: the bundle traverses.
- `test_static_routes.sh` asserts observable RIB effects per test (route-add counts by source, reload notices, route removals) with deadline polls instead of `kill -0`, on a dynamic port.

## Named follow-up

Configs that omit `storage:` entirely still panic by design in feature-reduced builds, which also keeps most config tests from running under `--no-default-features`. Surfacing the missing-default as a `Config::load` error instead of a panic would be the deeper fix; it changes the load API's contract, so it is left as its own decision.

### DIFF
--- a/bpa-server/docs/TODO.md
+++ b/bpa-server/docs/TODO.md
@@ -1,0 +1,20 @@
+# bpa-server TODO
+
+## Missing storage default: report as a `Config::load` error instead of panicking
+
+### Background
+
+`MetadataStorageConfig::default()` and `BundleStorageConfig::default()` (`src/config/storage.rs`) panic in builds compiled without the feature that provides their default backend (`sqlite-storage` for metadata, `localdisk-storage` for bundle data). The stance is deliberate: an unconfigured node must never silently degrade to a non-persistent memory backend, and the panic is pinned by `empty_config_has_defaults` in `src/config/mod.rs`, which flips to `should_panic(expected = "built without the")` under those feature combinations.
+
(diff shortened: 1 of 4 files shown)


ricktaylor (approved):
LGTM!!

PR #700, opened 2026-08-24
fix(tools): treat a failed capability probe as unprivileged

sylvain-pierrot:
# Fix the privilege probe and test the ping arithmetic

## Background

`tools/lib/privilege.rs` decides whether the CLI runs with elevated privileges. Its test asserted the CI runner's privileges rather than any code behavior: it failed on any unprivileged macOS or BSD machine, and passed on Linux only because of a bug in the function under test.

## What this changes and why

- **The bug:** the Linux capability probe used `caps::has_cap(...).is_ok()`, so a *failed probe* counted as privileged. It now treats a failed probe as unprivileged (`unwrap_or(false)`). The replacement test compares the function against platform ground truth, so it passes on privileged and unprivileged runners alike and fails exactly if the `is_ok()` regression returns.
- **Ping math, previously untested pure functions:** `build_payload`'s exact-target-size search is exercised across a 300-size sweep spanning the CBOR length-field boundaries, under a watchdog that turns a non-terminating binary-search mutant into a failure rather than a hang; `Statistics` (min/max under out-of-order recording, average, an exact stddev case, loss percentage, and the empty/single-sample edges) and `same_endpoint`'s RFC 9758 two-and-three-element ipn collapsing (with directly constructed legacy EIDs, since text never parses to the legacy form) are pinned.

## Verification

`cargo test -p hardy-tools`, clippy `-D warnings`, and fmt are clean; the privilege test passes as an unprivileged user here and is written to pass for root as well.

### DIFF
--- a/tools/lib/privilege.rs
+++ b/tools/lib/privilege.rs
@@ -21,13 +21,10 @@ pub fn has_required_privileges() -> bool {
         return true;
     }
 
-    // As a fallback, check if the user is running as root (EUID 0).
-    // A root user implicitly has all capabilities.
-    // Note: The `caps` crate can also check this, but a direct libc call is also common.
-    // We'll stick to the `caps` API for consistency here.
-    if caps::has_cap(None, CapSet::Effective, Capability::CAP_SETFCAP).is_ok() {
-        // A reasonable proxy for checking if the user is root, as only root can set file capabilities.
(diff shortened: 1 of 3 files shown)


ricktaylor:
I don't think this code is used anywhere :)

ricktaylor (approved):
LGTM!!

PR #701, opened 2026-08-24
test(tvr): run the e2e script against the current config schema

sylvain-pierrot:
# Revive the TVR e2e script: it has been silently dead

## Background

`tvr/tests/test_tvr.sh` stands up two BPA nodes, runs the Time-Variant Routing agent against a contact plan, and checks that routes appear and disappear on schedule. It configured the tcpclv4 CLAs with the removed `address` key, which the strict config schema refuses, so the nodes never started and every test in the script has been failing since that key was renamed; `set -e` additionally killed the script at the first failure, so its summary and result accumulation were dead code.

## What this changes and why

- **The nodes start again**: the CLA configs use the current `listeners` schema, verified against the bpa-server config code.
- **Failures are captured, not fatal**: each test runs under a condition so the per-test results and the summary always render, and the script still exits nonzero when any test failed. Unset results display as failures rather than crashing the summary.
- **No fixed sleeps, no fixed ports**: about 25 seconds of unconditional sleeping is replaced by readiness polls with deadlines (port probes and log-pattern waits that fail fast if a process dies), and ports are allocated dynamically so concurrent runs cannot collide.
- **The tests assert TVR's actual effect**: routes from the contact plan are asserted to land in and leave node 1's RIB via route-add/route-remove log polls, so a TVR that installs nothing now fails the suite.
- The config env-var tests get an RAII guard so a panicking assertion cannot leak `HARDY_TVR_*` into the other `#[serial]` tests.

## Verified and named limitations

Runs green end to end in a sandbox for TESTs 1 through 4; the gRPC-driven TESTs 5 through 10 skip cleanly where `grpcurl` is unavailable (their log patterns are verified against the TVR server source). Routing the ping's data path through node 1 is not achievable with the current config surface, because the tcpclv4 config has no static peer mapping for node 1 to learn node 2's socket address: the RIB-level assertions stand in for it, and a data-path assertion is a follow-up for when the config grows that capability.

### DIFF
--- a/tvr/src/config.rs
+++ b/tvr/src/config.rs
@@ -180,6 +180,9 @@ mod tests {
 
     // Sets env vars for its lifetime and removes them on drop, so a
     // panicking assertion cannot leak them into the other #[serial] tests.
+    // Removal (rather than restoring a prior value) is deliberate: an
+    // ambient HARDY_TVR_* would corrupt the unguarded config tests anyway,
+    // so clearing it is the safer disposition.
     struct EnvGuard(Vec<&'static str>);
 
     impl EnvGuard {
--- a/tvr/src/config.rs
+++ b/tvr/src/config.rs
@@ -178,6 +178,27 @@ mod tests {
         Config::load(Some(path)).unwrap()
     }
 
+    // Sets env vars for its lifetime and removes them on drop, so a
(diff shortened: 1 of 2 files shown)


ricktaylor (approved):
LGTM

PR #705, opened 2026-08-24
test(bpa): synchronize the channel tests on state transitions, not time

ricktaylor:
test_hysteresis_recovery ordered operations with polling loops, a 200ms quiesce window, and >=-under-deadline assertions - the shapes the test style guide bans, and the source of the recurring CI flake. The channel state machine gains a #[cfg(test)] transition Notify (the guide's hook for otherwise-unobservable private transitions; notify_one's permit semantics make check-then-await race-free), wait_for_state becomes event-driven, the drain runs to the exact expected count, and the re-open wait drains duplicate deliveries while awaiting the transition. Every timeout that remains only bounds a regression.


sylvain-pierrot (approved):
LGTM!!

PR #710, opened 2026-09-01
Fix/bpa dispatch redelivery

ricktaylor:
# Rationalise the dispatch/delivery egress state machine

This branch sits directly on `main`: the diff is exactly this PR's 16 commits — the six-commit state-machine rationalisation, and the ten commits that execute a three-pass adversarial deep review of it (35 findings, all dispositioned; the round-2 re-review verdict was "resolves the review, nothing blocks a merge").

## The bug that started it

The storage channel's queue poller could re-deliver a bundle that was already mid-processing: an in-flight `Dispatching` bundle recovered by a poll cycle became a second concurrent copy, which for local delivery meant duplicate concurrent `on_deliver` calls into the same service — unbounded under a congested channel. The fix direction is a queued-vs-in-flight split, and following it all the way through turned up the same missing distinction at every hand-off in the pipeline.

## The state machine

Local delivery is rebuilt to mirror forwarding rather than diverge from it. Queued dispatch work holds the new `DispatchPending` and is *claimed* to `Dispatching` by its consumer, so the poller only recovers genuinely queued work and duplicate copies lose the claim and drop. Each registered service gets a per-service queue in the new `DeliverPending` (the analogue of `ForwardPending`); its consumer claims each bundle to `DeliveryAckPending` before `on_deliver`, so deliveries serialise per service, every delivery exit resolves its claim, and delivery stops consuming dispatch pool slots. Unregistering sweeps the service's queue to `WaitingForService` (new `MetadataStorage::reset_service_queue`); restart re-parks the pending states and re-registration recovers them.

The other fixes ride the same claim discipline, each with its forwarding-side twin:

- **Missed re-dispatch**: a routing/registration event arriving while a bundle was in flight (claimed `*AckPending`, invisible to that event's poll) was lost when the later failure parked it. Every park now re-checks the routing table against the snapshot taken when the flight began and re-enters dispatch once if it changed — bounded, so a deterministic failure cannot spin. A re-dispatch re-enters from the persisted representation; filter and hop-count rewrites are attempt-scoped.
- **Two cold-start races** that stranded recovered bundles: the queue consumer starting before the CLA registry was wired, and recovery running concurrently with live registrations (mistaking a fresh forward for a stale checkpoint). The consumer now starts after wiring, `start()` awaits recovery to completion, and recovery's resets are conditional swaps. Builder-configured CLAs and services now also activate only after recovery, from `Bpa::start` — the quiescent-store contract holds on the server's own configuration path, not just dynamic registrations.
- **Stale-snapshot parks**: `forward_bundle`'s NoNeighbour and synchronous-failure arms, and two delivery exits (Deliver-filter error, no-key BPSec payload), parked from stale snapshots or with no recovery path; all now use conditional swaps, and a bundle for an unregistered local service parks as `WaitingForService`, never `Waiting`.
- **Truthful hand-offs across expiry**: once a bundle is mid-`on_deliver` or owned by a CLA transfer, the reaper defers rather than races it — a consumed bundle reports delivered and a completed transfer reports its outcome, instead of a contradictory "deleted: lifetime expired". Deferral does not launder expiry; dispatch and (now) the forward path both carry pre-claim expiry checkpoints, so an expired bundle is resolved as `LifetimeExpired` before any hand-off commences — it is never transmitted.
- **Deferred CLA transfer outcomes**: `ForwardBundleResult::Accepted` lets a reliable CLA take ownership and report `Completed`/`Failed` later via `Sink::transfer_outcome`, retained in `ForwardAckPending`. Resolution is arbitrated by the status-conditioned `MetadataStorage::swap_status` and its terminal `tombstone_if`, so an outcome racing a peer-loss sweep or a duplicate of itself is ignored.

## The egress queue model

The queue model is totalised: an egress policy classifies every bundle, so `queue_count()` is `NonZeroU32`, the queue index is a plain `u32` (out-of-range clamps to queue 0, logged), and `Option` survives only where it means something — the lane directive, where `None` is "next free lane", so the null policy imposes no lane constraint on a multi-lane CLA. The per-peer controller owns its queue mapping (`classify` deleted, `queue_for()` on the controller), taking the last shared policy state off the per-forward hot path. The trait vocabulary lands as the policy redesign names it: `EgressPolicy` → `FlowControllerFactory`, `EgressController` → `FlowController`. This PR also ports the #667 salvage: the null policy tolerates declared CLA lanes, and `ServiceRegistry::register` spawns its post-registration poll so a bridge-shaped sink cannot deadlock registration against its own announcements.

## Review hardening: invariants become compiler facts

The deep review's recurring theme was invariants maintained by comment discipline; these commits turn the load-bearing ones structural:

- **Reaper starvation, structurally impossible** (the review's one reproduced production bug): expired-but-unresolved hand-offs at the head of the expiry order could pin the reaper's refill page forever — one stalled peer stopped all expiry processing node-wide, an unbounded storage leak. **BREAKING:** `MetadataStorage::poll_expiry` loses its `limit`: backends are a status-blind ordered scan the *consumer* terminates, and the refill counts only accepted rows against its budget (deferral routed through the new exhaustive `BundleStatus::defers_expiry()`), so a skipped row costs nothing. Pinned by an end-to-end starvation regression test.
- **Every hand-off exit resolves its claim, by construction**: the offer bodies return a `#[must_use] OfferOutcome` consumed by a single resolver — a bare `return` inside an offer no longer compiles, and the three duplicated terminal blocks collapse into one.
- **No unconditional status writes remain.** **BREAKING:** `MetadataStorage::update_status` is deleted; every transition is a CAS. **BREAKING:** `Bundle` no longer implements serde — persistence deals in `StoredBundle`, whose only exit demands the status, so a backend cannot silently default a deserialized record to `New`.
- **By-construction shapes elsewhere**: `Service` is constructed complete (queue and canonical EID are constructor arguments; the dispatcher stops re-deriving the park key); peer ids are RAII `Reservation` guards; `EgressQueueSet` makes the always-present next-free queue a type-level fact; `unregister_cla` closes peer queues before sweeping (a racing forward bounces instead of stranding); sqlite's bulk SQL binds status codes through the codec (no literals; the on-disk numbering is frozen by a pin test) and postgres gains the `dispatch_pending` partial index its poller was missing.
- **Fail-fast storage**: a metadata-backend write failure can no longer masquerade as `DuplicateBundle` — the store fail-stops on backend error, matching every other storage seat, pinned by a `#[should_panic]` contract test. Dispatch claims moved into the pooled task, so throughput is no longer capped at one storage round trip per bundle.
- **Round-2 residue**: a registration losing the duplicate check now closes the delivery channel it opened (the channel's poller otherwise parked until store shutdown), and sqlite's `expiry` text is UTC-normalized at the write sites so the keyset cursor's lexicographic order is chronological by construction.

## Storage and compatibility

Backends gain the new statuses (SQLite codes 7/8/9; PostgreSQL migrations `0004_dispatch_delivery_statuses`, `0005_deliver_pending_index`, `0006_dispatch_pending_index`) and the shared metadata conformance suite pins the claim/park/reset/scan semantics across all three. **BREAKING** (full bullets in `bpa/CHANGELOG.md`): `Bpa::start` is async and runs recovery to completion; the policy trait overhaul; `poll_expiry`'s new contract; `update_status` removed; `Bundle` serde removed in favour of `StoredBundle`; `BundleStatus` serde derives removed; `services::registry` is no longer public.

## Verification

New event-driven rigs: dispatch-queue claim/duplicate/recovery interleavings, the forward/delivery expiry mirrors, the reaper starvation regression, the forward expiry checkpoint (proven to fail with the checkpoint removed), close-before-sweep unregistration (proven to fail on the old order), and the storage-suite extensions. Workspace `fmt --check`, `clippy --locked -D warnings`, and the full `--all-features` suite are green at this tip; CI must confirm the postgres suite against a live server (migrations 0004–0006).


sylvain-pierrot (approved):
LGTM!!

PR #716, opened 2026-09-05
Fix/creation timestamp uniqueness

ricktaylor:
## Unique-by-construction bundle ids — monotonic `CreationTimestamp::now`

`CreationTimestamp::now` derived its sequence number from the wall clock's nanoseconds (`nanosecond() % 1_000_000`), which makes bundle-id uniqueness probabilistic: two reads landing on the same instant produce identical `(time, sequence)` pairs. That happens in practice through two doors — concurrent issuers on separate cores reading the clock simultaneously, and coarse clocks (virtualised guests, embedded RTOS timers) where *sequential* calls within one tick collide systematically. `CLOCK_REALTIME` is also not monotonic: an NTP step backwards re-issues pairs from a millisecond that was already used. Callers papered over this with a rebuild-and-retry loop on `DuplicateBundle`.

`now()` now issues from a single packed atomic (`dtn_millisecs << 20 | sequence`), taking `max(wall_clock_candidate, last + 1)` per call: every pair is strictly greater than any pair this process has issued, bundles created in the same millisecond take ascending sequence numbers (the RFC 9171 §4.2.7 shape), concurrent issuers serialize through the CAS, and a clock stepping backwards never re-issues an earlier pair — it just continues ascending in the stale millisecond until the clock catches up. Issuing more than 2²⁰ pairs in one millisecond rolls the excess into the next millisecond's pairs, unique and self-correcting; 44 bits of DTN milliseconds last until the year 2557.

With uniqueness guaranteed by the Builder, `local_dispatch` drops its rebuild-and-retry loop: `services::Error::DuplicateBundle` from an application send now only means a collision with a pre-restart bundle. Uniqueness across process restarts still rests on the wall clock moving forward, per RFC 9171 §4.2.7, and the store's atomic insert refusal remains the backstop for exactly that case.

Two new integration tests pin the guarantee: strict sequential monotonicity, and cross-thread all-unique issuance — properties the atomic provides and clock resolution cannot.

## `portable-atomic` at the `no_std` RMW seats

The second commit converts `hardy-bpa`'s two production atomics — the service-number allocator (`AtomicU32::fetch_add`) and the storage channel state machine (`AtomicUsize` CAS) — from `core::sync::atomic` to `portable-atomic`, the source `hardy-bpv7` already uses. `core::sync::atomic` lacks `AtomicU64` without `target_has_atomic = "64"`, and load/store-only targets (thumbv6-class, riscv32 without the A extension) have no read-modify-write operations at any width; `portable-atomic` compiles to core atomics where they exist and polyfills them elsewhere. The crate's `std` and `critical-section` features forward to `portable-atomic`'s, mirroring `hardy-bpv7`'s wiring. No API change.

Gates: `fmt --check`, workspace `clippy --locked -D warnings`, and the full `--all-features` workspace suite are green.


bpa/src/dispatcher/local.rs
```diff
@@ -34,51 +34,47 @@ impl Dispatcher {
...
             }
+        }
 
-            let (raw, data) = builder
-                .with_payload(alloc::borrow::Cow::Borrowed(&payload))
-                .build(hardy_bpv7::creation_timestamp::CreationTimestamp::now())
-                .map_err(|e| services::Error::Internal(e.into()))?;
+        let (raw, data) = builder
+            .with_payload(alloc::borrow::Cow::Borrowed(&payload))
+            .build(hardy_bpv7::creation_timestamp::CreationTimestamp::now())
```
sylvain-pierrot: Why do we need `CreationTimestamp` as a parameter to `build()`? This could be the default behavior?

ricktaylor: Because in clockless systems you need to create your own CreationTimestamp by hand - so the builder cannot default to "use the system clock"

sylvain-pierrot: I meant, could you inline `now()` into the `build()` as a default behavior, and otherwise add a method to the builder to specify another time?

ricktaylor: Meh - I think that's a style issue I don't want to change now

sylvain-pierrot (approved):
LGTM!

PR #720, opened 2026-09-09
fix: block-type alias canonicalization; sqlite recovery of tombstoned orphans

ricktaylor:
Three unrelated fault fixes against main, surfaced by the refactor/filters review round (F-37 via prior-art comparison, F-36 promoted from out-of-scope debt).

## bpv7: `Unrecognised(v)` aliases bypassed every block-type policy

`Type` had two spellings for one wire value: `Unrecognised(v)` encodes as the raw code `v`, so a hand-built `Unrecognised(11)` is a Block Integrity Block to the next parser while matching no named variant locally. `Editor::push_block`'s refusals (primary, BIB/BCB — the Signer/Encryptor gate) and singleton rules all matched variants, so an alias walked past all of them: 11/12 emitted wire-valid fake security blocks, 0 bypassed the primary refusals, 1/6/7/10 bypassed the duplicate rules. `Builder::add_extension_block` had the same hole. Everything downstream of parse already assumes "named variant ⟺ known code" (parse folds through `From<u64>`); these were the only two doors where an unfolded `Type` could enter.

The new `block::Type::canonicalize()` folds an alias of a known code back to its named variant — definitionally the composition of the two existing `From` impls — and both doors canonicalize before their unchanged variant matches. A sweep of the other catch-all numeric enums found no further instances: `CrcType` aliases already fail closed, `ReasonCode::Unassigned(n)` is wire-identical to its named form, and the BPSec context/algorithm enums have no public construction doors.

## sqlite-storage: tombstoned orphans crash-looped startup recovery

A tombstone nulls the row's columns but keeps the row (duplicate suppression); a crash between tombstone and data deletion strands a blob whose id resolves to that row. `confirm_exists` was the one tombstone-aware read without the `bundle IS NOT NULL` predicate, so recovery's row mapper hit `InvalidColumnType` on the NULL column and `trace_expect`'d — a startup panic recurring every boot, cleared only by manual database surgery. postgres was unaffected (its tombstone deletes the row).

With the predicate (the file's own nine-occurrence tombstone idiom), recovery self-heals: `Ok(None)` routes the blob through the orphan path, the re-ingress reports it as a duplicate of the tombstone, and the stranded data is deleted. META-05 in the shared storage suite now pins the contract per persistent backend.

## sqlite-storage: the poll_pending drain was O(B²) under backlog

The status-keyed poll had no index covering its ORDER BY, so every 16-row drain page top-K-sorted the entire matching backlog through a temp B-tree — quadratic row visits to drain a backlog, on the node-wide dispatch queue and every per-service delivery queue, engaging exactly in the overload regime the storage poller exists for. Migration `02_poll_pending_index` adds a covering index (the two superseded indexes are strict prefixes of it, so every other status-keyed query keeps an equal-or-better plan); the poll SQL is hoisted to a const so the new plan pin test EXPLAINs the exact production query and fails if it ever sorts out of index. The temp B-tree and its elimination were both reproduced against the real schema during review verification (review F-07). The column order deliberately keeps `received_at` ahead of `status_param3` so the index also serves the forwarding-queue poll shape arriving with the egress rework.


sylvain-pierrot (approved):
LGTM!

PR #721, opened 2026-09-09
feat(bpa)!: the CLA acceptance contract — explicit dispatch verdicts, negotiated size caps, persisted next-hop

ricktaylor:
# The CLA acceptance contract

First of the refactor train, directly on `main`; the filter subsystem rework (#712) sits on this branch. This work was originally interleaved with the filters branch and has been split out: it changes the CLA/BPA contract, not the filter system, and its fixes and review findings deserve their own review.

## The verdict

`cla::Sink::dispatch` returns the new `#[must_use]` `cla::Acceptance`: `Accepted` means the CLA may acknowledge the transfer — internal disposal of an invalid or duplicate bundle is still acceptance, since refusing content that cannot change would invite resends of identical bytes forever, so the first BPA to detect a problem accepts and terminates the bundle — and `Refused` (a truncated stream, or the size cap) means it must not, leaving the peer responsible. `Err` no longer carries a verdict: it is reserved for sink faults, and a CLA's own unregistration mid-stream surfaces as `Err(Disconnected)`. The bundle stream itself carries no verdict either — the BPA may stop reading once it has decided, so a producer that sees `SendError` stops pushing and takes the verdict from the return value alone.

## The negotiated size cap

To keep deterministic refusals from ever being provoked, the cap is negotiated at registration: `BpaBuilder::cla` and `register_cla` carry the CLA's declared transport limit, and `Cla::on_register` receives the effective cap — the BPA's configured `max_bundle_size` (now `NonZeroU64`, dispatch-enforced) folded with the declaration. Both directions travel the gRPC wire (`RegisterClaRequest`/`RegisterClaResponse.max_bundle_size`, `DispatchVerdict`); absent values from an older BPA read as accepted / effectively-uncapped. tcpclv4 declares its transfer MRU, advertises the effective cap to peers, and answers a refusal with XFER_REFUSE (keeping the session — teardown is for faults only); file-cla skips over-cap files unread; BIBE pre-checks the encapsulated outer before dispatch.

## The forwarding record

`BundleStatus::ForwardPending` carries the resolved adjacency in a new `next_hop: Eid` field: the routing decision rides the queue-assignment record, so the egress channels' at-least-once storage recovery re-delivers it intact (previously the adjacency lived in a never-persisted transient and a recovered bundle forgot its next hop). Queue polls match on queue identity via the new `BundleStatus::same_queue`; ownership swaps keep full equality. `local.rs` is split into `originate.rs` and `deliver.rs` along the same seam.

## Review fixes riding along (each with its regression test)

The deep review of the original combined branch found three places where this contract's own adaptations answered per-bundle outcomes with the wrong signal class:
- **bibe** reported an over-cap outer and a BPA refusal as `NoNeighbour` — a link-scoped signal that sweeps the entire peer queue. They are now `Err(PayloadTooLarge)` and the typed `bibe::Error::Refused`; `NoNeighbour` is reserved for a genuinely disconnected sink.
- **file-cla** left refused/failed files "for the next scan" with no scan in existence. The watcher now sweeps the outbox at startup (watch-then-sweep), matches rename-into events (atomic spooling works), quarantines deterministic refusals to `outbox/refused/`, and leaves transient failures for the next sweep.
- **the proto CLA client** decoded unknown `DispatchVerdict` values as Accepted via the prost getter's default collapse — silent bundle loss on any future verdict. The raw field is matched; unknown values fail closed.

## Interim notes (removed by #712)

This branch runs against the existing filter engine, so three deliberate redundancies exist until the filter rework lands: the new `BpaBuilder::primary_block_integrity`/`bundle_age_required` knobs coexist with the still-registered `Rfc9171ValidityFilter` (identical verdicts, benign); the new built-in per-hop legacy re-encode stage (`BpaBuilder::ipn_legacy_peers`) defaults empty and inert while `hardy-ipn-legacy-filter` remains the configured path; and `metadata.next_hop` survives as a shim populated from the persisted status for the legacy egress `WriteFilter` API.

## Verification

Full workspace `fmt --check`, `clippy --locked -D warnings`, and `--all-features` suites green at this tip (environment-only exceptions: postgres/s3 need live backends, tcpclv4 `[::1]` needs IPv6). New tests: the bibe refusal pair, the file-cla sweep/quarantine pair, and the fail-closed verdict decode.


sylvain-pierrot (dismissed):
LGTM

PR #722, opened 2026-09-09
refactor(bpa)!: rename routing::RoutingSink to routing::Sink

ricktaylor:
The registry sink traits are cla::Sink and services::Sink; the routing agent sink alone carried its module name in the type. The module path already says whose sink it is, so the prefix is noise — the trait is now routing::Sink, matching its siblings.

One mechanical commit: the trait, its re-export, and all references across bpa, bpa-server, tvr, and proto. Test mocks keep their local RoutingSink wrapper names where they read naturally; the one genuine collision (the registry's implementation struct is itself named Sink) keeps the trait path-qualified at the impl. BREAKING for external routing agents: a one-token rename at the impl and on_register signature.

No behaviour change; no filters/refactor-train dependency — this lands off main and the train absorbs it at the next cascade.


ricktaylor:
On consideration - this is the wrong move. Renaming cla::Sink to cla::ClaSink is probably more correct

PR #723, opened 2026-09-09
refactor(bpa,proto)!: door errors carry u64 sizes

ricktaylor:
The size-carrying door errors move to the wire-size domain: services::Error and cla::Error PayloadTooLarge / PayloadUnderrun carry { size: u64, ... } (were usize), so a declared size beyond a 32-bit target's address space reports its true value. The in-memory accumulator errors (stream::ConcatError / BufferError) deliberately keep usize: they measure buffered bytes, not wire declarations, and widen at the door conversions instead. buffer_stream tidies to match (the declared length converts once, under its own name). The ingress door's refusals, the raw originate door, and the gRPC client's pre-flight and interim buffering sites widen at the same boundary.


sylvain-pierrot (approved):
LGTM

PR #724, opened 2026-09-09
feat(bpa,proto,tcpclv4,file-cla,bibe)!: registration negotiates the bundle size cap

ricktaylor:
The BPA has enforced max_bundle_size since the streamed doors landed, but CLAs could only discover it by tripping over it. Registration now negotiates it: the CLA declares its transport limit, the BPA folds it with its configured cap, and Cla::on_register delivers the effective value for the CLA to propagate into its own transfer advertising — an over-cap transfer is never offered, because rejection is deterministic and a peer that keeps offering the same bundle loops forever. tcpclv4 clamps its advertised MRU to the effective cap; file-cla skips over-cap outbox files unread; BIBE pre-checks the encapsulated outer before dispatch. Both directions travel the gRPC registration wire, with absent values from older peers degrading safely. BREAKING: register_cla/BpaBuilder::cla gain the declared-limit parameter, on_register gains the effective cap, and BpaBuilder::max_bundle_size takes NonZeroU64.


sylvain-pierrot (approved):
LGTM

PR #727, opened 2026-09-11
Fix/storage replace tombstone

sylvain-pierrot:
Extracted from the #703 review round ([pr703-additions §2](https://github.com/ricktaylor/hardy/blob/main/references/reviews/pr703-additions.md)) as a standalone fix so it lands ahead of that PR; #703 sheds the overlapping doc renumbering on its next rebase.

The bug. MetadataStorage::replace is an unconditional write, so it races tombstone, the peer-loss sweeps, and the expiry reaper — and both SQL backends mishandled the lost race. sqlite: deletion nulls the row's columns in place, so the unqualified UPDATE wrote the caller's snapshot straight back in, resurrecting the deleted bundle and forfeiting duplicate protection. postgres: deletion drops the metadata row, so the update matched nothing and returned RowNotFound, which Store::update_metadata escalates to a panic — any metadata write losing an expiry race took the node down.

The fix. sqlite's UPDATE gains the same bundle IS NOT NULL predicate every other tombstone-aware query uses (a lost race now logs at debug, matching tombstone); postgres stops treating zero rows as an error. Both now give the outcome the in-memory backend already enforced ("expiry wins"), and the trait rustdoc documents the contract: a tombstoned bundle is never resurrected, and a write that matches nothing is not an error. The postgres half is included by review ruling (2026-09-11), superseding the earlier deferral — the refactor train that deletes replace hasn't landed, and until it does a lost race is a production panic.

Test. New harness test META-17 (insert → tombstone → mutate snapshot → replace → get() still None, re-insert() still refused) runs against all three registered backends and fails on both unfixed SQL backends. The five doc-file renumberings correct counts that were already stale (META-15/16 were never propagated to the backend docs).


ricktaylor (approved):
LGTM

PR #728, opened 2026-09-11
Fix: dispatch outbox files that predate the watch

sylvain-pierrot:
Extracted from #703 (the storage/CLA test-overhaul branch), where this fix first surfaced, so it can land ahead of that PR. #703 will shed these files on its next rebase.

The bug
The watcher only reacts to Create(File) filesystem events, which cover files created after the watch is installed. A bundle queued in the outbox while the CLA was not running — the removable-media use case the design doc promises, or simply a restart with queued files — was never dispatched: it sat in the directory indefinitely until something touched it again.

The fix
watcher_task now scans the outbox once, immediately after installing the watch, and feeds any file already present into the same channel the event path uses. The ordering matters: with the watch installed before the scan, a file arriving mid-scan is seen by at least one of the two paths, and the forwarder tolerates the resulting duplicate (the second read finds the file already dispatched and deleted, logs, and moves on). Scan errors are logged and non-fatal.

Deliberate difference from the version in #703
The scan uses metadata() (which follows symlinks) rather than DirEntry::file_type() (which does not). inotify reports the creation of a symlink as an ordinary Create(File), so the event path dispatches symlinked files; a scan that skipped them would leave the two paths disagreeing about the same directory entry. This is pinned by the new test_outbox_pickup_preexisting_symlink test and a design-doc paragraph, both of which exist only on this branch — when #703 rebases over this, the scan here is the version to keep.

Also carried
The file-cla test coverage from #703 rides along unchanged (tests/cla.rs forward-path suite, tests/watcher.rs, canonicalize_path unit tests, tempfile dev-dependency): the watcher tests are what pin this fix, and the rest re-sequences already-reviewed #703 content so that branch shrinks. The event-path test uses a two-phase marker (the marker's dispatch proves the startup scan already ran) so the scan cannot mask an event-path regression; waits are bounded by a generous deadline that only bounds a regression, since real filesystem events offer no synchronous hook.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2101,6 +2101,7 @@ dependencies = [
  "notify",
  "notify-debouncer-full",
  "serde",
+ "tempfile",
  "thiserror",
  "tokio",
  "tokio-util",
--- a/file-cla/CHANGELOG.md
+++ b/file-cla/CHANGELOG.md
@@ -9,6 +9,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 ### Changed
 - Adapted to the `hardy-bpa` deferred transfer-outcome CLA contract (new `Cla::forward` signature). Behaviour is unchanged: forwards remain terminal.
 
+### Fixed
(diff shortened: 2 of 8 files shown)


ricktaylor (approved):
LGTM

